PR #11, opened 2025-07-16
Building aars script

sKrzysieK:


SuperProject/android/app/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.CAMERA" />
```
mczernek, line 4: This shouldn't be necessary, package (aar) should be able to add that (according to https://developer.android.com/build/manage-manifests)

PR #16, opened 2025-11-03
move plugin do packages/client

radoslawrolka:
- move plugin do packages/client
- added support for packges with pattern: `<@.../....>: <version>`
- added expo plugin


build_aar.sh
```diff
@@ -28,13 +28,14 @@ create_temp_rn_project() {
     local RN_VERSION=$1
     local TEMP_DIR=$2
     local IS_EXPO=$3
+    rm -rf "$TEMP_DIR" 2>/dev/null
 
     if [ "$IS_EXPO" = true ]; then
```
WoLewicki, line 33: We don't care about the expo in the end so we can remove it. But please do it in follow-up

build_aar.sh
```diff
@@ -97,6 +87,19 @@ for IS_EXPO_PROJECTS in false true; do
                 pushd "$TEMP_PROJECT_DIR" > /dev/null
                 install_dependencies "$PACKAGE_NAME" "$VERSION" "$IS_EXPO_PROJECTS"
 
+                # Copy LICENSE file if exists
+                LICENSE_PATH="node_modules/$PACKAGE_NAME/$(ls "node_modules/$PACKAGE_NAME" | grep -i 'License*')"
+                if [[ -f "$LICENSE_PATH" ]]; then
```
WoLewicki: Do we want to check what we exactly found in any way? There's a change we find some weird implementation file or whatever named that way yeah?

example.rnrepo.config.json
```diff
@@ -1,4 +1,3 @@
 {
-    "disableRnRepo": false,
     "denyList": ["react-native-reanimated", "react-native-svg"]
 }
```
WoLewicki: newline

.gitignore
```diff
@@ -7,4 +7,7 @@ build
 dist
 *.aar
 *.pom
+AARS/
```
WoLewicki: are there any items generated here?

withBuildlePlugin.js
```diff
@@ -0,0 +1,55 @@
+const { withProjectBuildGradle, withAppBuildGradle } = require('@expo/config-plugins');
+
+const classpathRegex = /(classpath.*)/;
+const buildleClasspath = 'classpath("com.swmansion:buildle-plugin:1.0.6")';
```
WoLewicki: I think we should get the plugin version directly from somewhere and not hardcode it here right?

withBuildlePlugin.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+const withCustomBuildSettings = config => {
+    config = withClasspathDependency(config);
+    config = withMavenRepository(config);
+    config = withBuildlePluginApplication(config);
+    return config;
+};
+
+module.exports = withCustomBuildSettings;
```
WoLewicki: newline

withBuildlePlugin.js
```diff
@@ -0,0 +1,55 @@
+const { withProjectBuildGradle, withAppBuildGradle } = require('@expo/config-plugins');
```
WoLewicki, line 1: Where is this script used?

radoslawrolka: This script can be used inside expo project, so `npx expo prebuild -p android` wont delete our rnrepo setup in gradle files

withBuildlePlugin.js
radoslawrolka: done

withBuildlePlugin.js
radoslawrolka: Could be done with some .env, but maybe better would be pulling newest version?
```suggestion
const buildleClasspath = 'classpath("com.swmansion:buildle-plugin:+")';
```

.gitignore
radoslawrolka: during build_aars.sh, but removed it from .gitignore

example.rnrepo.config.json
radoslawrolka: done

build_aar.sh
radoslawrolka: removing license file from aar due to changes in project

PR #43, opened 2025-11-14
[FEATURE] Reduce size of reanimated Aars

radoslawrolka:
Prefab includes unstripped libraries, this change adds improved stripping resulting in 90% aars size reduction


packages/builder/prefab-reduce.gradle
```diff
@@ -0,0 +1,28 @@
+allprojects { project ->
+    if (project.name == "react-native-reanimated") {
```
kmagiera: isn't it ok to keep this for any library instead of filtering reanimated here?

radoslawrolka: done

PR #46, opened 2025-11-15
Better developer mode and linters

radoslawrolka:


packages/client/rnrepo-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
```diff
@@ -43,15 +44,8 @@ class PrebuildsPlugin : Plugin<Project> {
...
-                    if (project.hasProperty("RNREPO_USE_DEV_REPO") &&
-                        project.property("RNREPO_USE_DEV_REPO").toString().toBoolean()) {
-                        repo.name = REMOTE_REPO_NAME_DEV
-                        repo.url = URI(REMOTE_REPO_URL_DEV)
-                        project.logger.lifecycle("[RNRepo] Using DEV remote repository: $REMOTE_REPO_URL_DEV")
-                    } else {
-                        repo.name = REMOTE_REPO_NAME_PROD
-                        repo.url = URI(REMOTE_REPO_URL_PROD)
-                    }
+                    repo.name = REMOTE_REPO_NAME
```
kmagiera, line 46: is it important that the name is provided? you don't need to set the name when defining repo url in the repositories configuration. This would let us make the custom configuration a bit leaner avoiding one extra parameter from being specified.

packages/client/rnrepo-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
```diff
@@ -244,7 +241,7 @@ class PrebuildsPlugin : Plugin<Project> {
             return true
         }
 
-        val urlString = "https://packages.rnrepo.org/releases/org/rnrepo/public/${gradlePackageName}/${packageVersion}/${gradlePackageName}-${packageVersion}-rn${RNVersion}.aar"
+        val urlString = "${REMOTE_REPO_URL}/org/rnrepo/public/${gradlePackageName}/${packageVersion}/${gradlePackageName}-${packageVersion}-rn${RNVersion}.aar"
```
kmagiera: couldn't we check `repositories` configuration for the current project instead of hardcoding / expecting certain env variable to be properly set? Now the variable and maven repo configuration is coupled, if you provide the dev repo URL via env but have releases set in the maven config, the checks will be performed agains the wrong repo

packages/client/rnrepo-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
radoslawrolka: Repo name is mainly used for debugging, so that it is clear where maven gets the artifacts from, but the 'packages.rnrepo.org' url link itself is informative enough. Will remove the name

packages/client/rnrepo-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
radoslawrolka: done, searching in all project repos
```
        project.repositories.forEach { repo ->
            val urlString = "${repo.url}/org...
```

PR #83, opened 2025-11-27
[CHORE] Check for new architecture in PrebuildsPlugin 

radoslawrolka:
If old arch is detected or <=0.75 rn is detected then architecture condition is not passed

### DIFF
--- a/packages/client/rnrepo-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
+++ b/packages/client/rnrepo-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
@@ -258,8 +258,8 @@ class PrebuildsPlugin : Plugin<Project> {
         }
 
         // RN <= 75: Plugin supports only >= 0.76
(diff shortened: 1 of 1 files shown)


packages/client/rnrepo-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
```diff
@@ -202,16 +202,79 @@ class PrebuildsPlugin : Plugin<Project> {
...
+            logger.warn("Could not parse React Native version: $rnVersion, assuming New Arch is disabled")
+            return false
+        }
+
+        val major = versionParts[0].toIntOrNull() ?: 0
+        val minor = versionParts[1].toIntOrNull() ?: 0
+
+        // RN >= 0.82: New Arch is default
+        if (major > 0 || (major == 0 && minor >= 82)) {
+            logger.info("React Native $rnVersion >= 0.82, New Architecture is enabled by default")
```
WoLewicki, line 256: I'd even change it to `New Architecture is the only available one`

PR #150, opened 2025-12-05
Rolkrado/simplify expo plugin

radoslawrolka:
- simplify plugin implementation
- add timeout and concurrency group to CIs

### DIFF
--- a/.github/workflows/build-example-app.yml
+++ b/.github/workflows/build-example-app.yml
@@ -1,0 +1,81 @@
+name: Build Example App with Android Gradle Plugin
+
+run-name: Build Example App with Android Gradle Plugin
+
+on:
+  workflow_dispatch:
+    inputs:
+      maven_repository_type:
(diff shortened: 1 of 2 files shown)


packages/expo-config-plugin/src/withRNRepoPlugin.ts
```diff
@@ -1,27 +1,19 @@
-import { withProjectBuildGradle } from '@expo/config-plugins';
-import { withAppBuildGradle } from '@expo/config-plugins';
+import { withProjectBuildGradle, withAppBuildGradle } from '@expo/config-plugins';
```
WoLewicki, line 1: You mentioned that this didn't work earlier?

radoslawrolka: Tested locally with changed configuration and it worked, so I assume it will also work on published package

PR #168, opened 2025-12-10
[BUG] Fix building with wrong RN version

### LINKED ISSUE #167: [BUG] Libraries for non-existing RN versions can be build and publish
### 📋 Description

It appears that due to an issue in CLI's init scripts when we provided a non-existent RN version, the build workflow would still produce an artifact and publish it. As a consequence we have artifacts for 0.79.9 version now published while this version doesn't exists. The build script should fail when an invalid RN version is provided

### 🔄 Steps to Reproduce

1. go here: https://packages.rnrepo.org/releases/org/rnrepo/public/react-native-gesture-handler/2.28.0/

### 🤔 Expected Behavior

There should be no builds for RN 0.79.9

### 😲 Actual Behavior

There are builds for RN 0.79.9

### 📸 Screenshots / Logs

_No response_

### RNRepo version

0.1.0

### React Native version

0.79.9

### Gradle version

8.3

### Operating System

macos

### 📦 Related Libraries (if applicable)

_No response_

### ➕ Additional Information

_No response_

radoslawrolka:
## 📝 Description

Now the scheduler correctly fetches and matches React Native versions from npm to schedule only compatible builds. Also builder now validates the RN version before starting the build process.

## 🎯 Type of Change

Please delete options that are not relevant:

- [x] 🧪 Test addition/modification

## 🧪 Testing

Please describe the tests you ran to verify your changes. Provide instructions so we can reproduce.

- run scheduler tests
- run scheduler locally to check if it will schedule invalid rn build
  - requires adding invalid rn version to `react-native-versions.json`

## 📋 Additional Notes

- fixes #167

### DIFF
--- a/packages/scheduler/src/scheduler.ts
+++ b/packages/scheduler/src/scheduler.ts
@@ -44,16 +44,18 @@ export async function processLibrary(
         pkgMatcher,
         publishedAfterDate
       );
+      // If reactNativeMatcher is not set, accept any version
+      const reactNativeMatchingVersions = await findMatchingVersionsFromNPM(
+        'react-native',
+        reactNativeMatcher ?? '*'
+      ).then(versions => versions.map(v => v.version));
 
       for (const pkgVersionInfo of matchingVersions) {
         const pkgVersion = pkgVersionInfo.version;
 
         for (const rnVersion of rnVersions) {
-          // If reactNativeMatcher is not set, accept any version
-          if (
-            reactNativeMatcher &&
-            !matchesVersionPattern(rnVersion, reactNativeMatcher)
-          ) {
+          if (!matchesVersionPattern(rnVersion, reactNativeMatchingVersions)) {
+            console.log(`   ❌ Skipping RN ${rnVersion} - does not match reactNativeVersion criteria`);
             continue;
           }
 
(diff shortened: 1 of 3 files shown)


packages/scheduler/src/scheduler.ts
```diff
@@ -44,16 +44,18 @@ export async function processLibrary(
         pkgMatcher,
         publishedAfterDate
       );
+      // If reactNativeMatcher is not set, accept any version
+      const reactNativeMatchingVersions = await findMatchingVersionsFromNPM(
```
kmagiera, line 48: there's already call to `findMatchingVersionsFromNPM` above. Did you forget to delete it? Do we need them both?

kmagiera (approved):
Left one comment inline. Please address it before merging

packages/scheduler/src/scheduler.ts
radoslawrolka: There are 3 `findMatchingVersionsFromNPM` calls:
- one for worklets version (optional)
- one for react-native version
    - This is the added one
    - Invalid rn versions builds were scheduled due to lack of validation
- one for the builded package version

PR #172, opened 2025-12-11
Rolkrado/web new layout

radoslawrolka:
## 📝 Description

Zrzut ekranu 2025-12-11 o 11 37 47

## 🎯 Type of Change

- [x] 📚 Documentation update

## 🧪 Testing

**Steps to reproduce:**
1. `cd packages/website && bun run dev`


kmagiera (approved):
Looks good. The RN versions section still takes too much space. I think we should just use the same approach as with the library versions. 

Also, the column with library versions shouldn't be labeled as "Android versions" but "Library versions" instead

PR #175, opened 2025-12-11
[CI] Remove snapshot repo secret and use format

radoslawrolka:
## 📝 Description

[CI] Remove snapshot repo secret and use format

## 🎯 Type of Change

- [x] ♻️ Refactoring (no functional changes)


radoslawrolka:
⚠️ REQUIRES CHANGING GH SECRETS (do not forget)

kmagiera (approved):
This is ok, but maybe we should use "URL_BASE" suffix because for this change we will need to update the existing secret as it already contains "/releases" path

PR #178, opened 2025-12-12
Enhancemenets to Supported Libraries page

kmagiera:
Brings some enhancements and fixes to supported libraries page:
1) render table items in multiple-rows on mobile
2) add padding between beta notice block
3) don't show the number of libraries more than once
4) fix weird highlighting of version buttons
5) remove "benefits" from navbar and add "supported libraries" to footer
6) add library list mock used in dev mode (such that we don't need to have db credentials when developing)
7) fix navbar wrapping text in mid-width – we now use github icon instead of "view on github" in mid-width windows


kmagiera:
image
image
image

PR #187, opened 2025-12-16
[BUMP] Bump plugin to 0.2.0

radoslawrolka:
## 📝 Description

Bump client, expo-plugin versions and all readmes

## 🎯 Type of Change

- [x] 📚 Documentation update

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -56,7 +56,7 @@ body:
     id: rnrepo-version
     attributes:
       label: RNRepo version (and/or @rnrepo/expo-config-plugin version)
-      placeholder: "e.g., 0.1.0"
+      placeholder: "e.g., 0.2.0"
     validations:
       required: true
 
--- a/.github/workflows/build-example-app.yml
+++ b/.github/workflows/build-example-app.yml
@@ -17,7 +17,7 @@ on:
         description: 'RNRepo prebuilds-plugin version'
         required: false
         type: string
-        default: '0.1.0'
+        default: '0.2.0'
       libraries:
         description: 'libraries with versions to install, format: libraryName1@version1,libraryName2@version2'
         required: true
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ For standard React Native setups or when using Expo but managing your android fo
      }
      dependencies {
        ...
-   +   classpath("org.rnrepo.tools:prebuilds-plugin:0.1.0")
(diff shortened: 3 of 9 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -56,7 +56,7 @@ body:
     id: rnrepo-version
     attributes:
       label: RNRepo version (and/or @rnrepo/expo-config-plugin version)
-      placeholder: "e.g., 0.1.0"
+      placeholder: "e.g., 0.2.0"
```
WoLewicki, line 59: No need to change it really I guess?

.github/workflows/build-example-app.yml
```diff
@@ -17,7 +17,7 @@ on:
         description: 'RNRepo prebuilds-plugin version'
         required: false
         type: string
-        default: '0.1.0'
+        default: '0.2.0'
```
WoLewicki, line 20: Can we just take it from the package.json ?

README.md
```diff
@@ -72,7 +72,7 @@ For standard React Native setups or when using Expo but managing your android fo
      }
      dependencies {
        ...
-   +   classpath("org.rnrepo.tools:prebuilds-plugin:0.1.0")
+   +   classpath("org.rnrepo.tools:prebuilds-plugin:0.2.0")
```
WoLewicki, line 75: Here we can also read it from package.json?

bun.lock
```diff
@@ -59,7 +59,7 @@
     },
     "packages/expo-config-plugin": {
       "name": "@rnrepo/expo-config-plugin",
-      "version": "0.1.0-beta.1",
+      "version": "0.2.0-beta.0",
```
WoLewicki, line 62: Could you remind me why is is `beta` at all here?

.github/ISSUE_TEMPLATE/bug_report.yml
radoslawrolka: yes

.github/workflows/build-example-app.yml
radoslawrolka: it could be, but i wanted to have arbitrary env for testing purposes

bun.lock
radoslawrolka: Just a convention, due to the fact that rnrepo is still in beta phase.

README.md
radoslawrolka: I dont think reading version value from build.gradle is possible on the markdown side (without any additional scripts files, which imo would be overengineering this)

PR #189, opened 2025-12-17
[LIBS] Remove nitro libs for stability

radoslawrolka:
## 📝 Description

Remove libs based on nitro libs due to unstable cpp

- react-native-iap
  - removing
    - ```
      * What went wrong:
      Cannot locate tasks that match ':react-native-iap:bundleReleaseAar' as task 'bundleReleaseAar' is ambiguous in
      project ':react-native-iap'. Candidates are: 'bundleAmazonReleaseAar', 'bundleAmazonReleaseLocalLintAar', 
      'bundlePlayReleaseAar', 'bundlePlayReleaseLocalLintAar'.
      ```
  - includes nitro from 14.0.0, can support lower versions then
  - https://www.npmjs.com/package/react-native-iap?activeTab=versions
    - v12.16 (without nitro) have +10k weekly downloads, so worth supporting
- react-native-mmkv
  - includes nitro from v4, can support lower versions then
  - https://www.npmjs.com/package/react-native-mmkv?activeTab=versions
     - v3.x (without nitro) have +300k weekly downloads, so worth supporting
  - ``` 
      error: License (MIT AND BSD-3-Clause) is not allowed. 
      Allowed licenses are: MIT, Apache-2.0, BSD-3-Clause, BSD-2 Clause
     ```
- react-native-unistyles
  - removing
- react-native-nitro-sound
  - removing

## 🎯 Type of Change

- [x] 🧩 New library support

### DIFF
--- a/libraries.json
+++ b/libraries.json
@@ -489,6 +489,14 @@
       }
     ]
   },
+  "react-native-mmkv": {
+    "android": [
+      {
+        "versionMatcher": "<4.0.0",
+        "publishedAfterDate": "2025-01-01"
+      }
+    ]
+  },
   "react-native-edge-to-edge": {
     "android": [
       {


kmagiera:
mmkv 3.3.3 is the most frequently installed version now and released only 3 months ago. I think it should be safe for rnrepo to build versions <4 of mmkv

PR #195, opened 2025-12-19
[CI] Draft ios build yml for early merge

radoslawrolka:
## 📝 Description

This workflow does not work yet. To run CI from branches (testing) yml file must exist on main

## 🎯 Type of Change

### DIFF
--- a/.github/workflows/publish-library-ios.yml
+++ b/.github/workflows/publish-library-ios.yml
@@ -1,0 +1,103 @@
+name: Publish React Native Library for iOS
+
+run-name: Publish library for iOS from build workflow
+
+on:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-library-android.yml
```diff
@@ -30,6 +30,7 @@ on:
 jobs:
   build-library-android:
     runs-on: rnrepo-builder-ubuntu-latest
+    timeout-minutes: 60
```
WoLewicki, line 33: why is this in this PR?

radoslawrolka: to keep same configuration in ios and android build workflow

PR #198, opened 2025-12-23
Initial implementation for ios prebuild script and plugin for replacing dependencies

kmagiera:
This PR adds the initial version of iOS scripts for prebuilding library as frameworks and a way of incorporating these in the apps.

The way the build script work:
1) For build scripts we extract shared logic between setting up the project from android scripts
2) The actual build step builds the entire app project and then extract the library pod xcframework from outputs directory
3) Since xcframework is a directory, we use zip to compile it into a file artifact that uses the same naming convention as on android: library_name-version-rn{rn_version}-config.xcframework.zip
4) In addition, we append "config" which is either "release" or "debug"
5) For development purposes the frameworks are only build for iossimulators, we should enable full simulator + device builds later.

For testing you can run the ios build script like so:
```
bun run build-ios react-native-svg 15.15.1 0.82.1 debug /tmp/builddir
```

After the build is complete, you should find the build artifact under `outputs/` directory in the build folder.

Now, here's how the plugin works:
1) We implement a cocoapods plugin and in order to use it in the app you need to `bun run install` in plugin directory and then add `plugin 'cocoapods-rnrepo'` in the app's Podfile. This isn't the final way for us to distribute the plugin, perhaps some local imports will work better (see discussion and future plans below)
2) The plugin works in `pod install` phase, and traverses the dependencies. It detects dependencies that are likely coming from autolinking and are react native third party libraries (don't live under `react-native/` and are linked as "development pods"). For detected libraries, we check whether the prebuilds exists by constructing the file name using the convention mentioned earlier. For development purposes we check `/tmp/rnrepo` folder first where we can place the test builds. Otherwise we check against maven repository location and try to download from there.
3) For detected libraries for which prebuilt exists, we perform a substitution by overwriting the pod dependency info. This makes it so that instead of the dependency to load files for compilation, it will use the pre-built framework instead.
4) Since we only know the build variant in build time and not pod install time, we use a similar approach to RN prebuilds and rely on a concrete path where the framwework file is going to be placed. Then we use a build script that would use symlink to alternate between linking to Release or Debug build for that fixed location.
5) In order to link xcframework in pods, we need to provide a location. We use the `node_modules/library_name` location to unzip the build artifact to. We create a `.rnrepo-cache` directory in each individual library location. This also helps us avoid re-downloading and re-unzipping the artifacts in future runs.
6) The symlink is also placed under that same `.rnrepo-cache` location under `.rnrepo-cache/Current` while `Debug` and `Release` versions have their individual folders.

After the `pod install` phase is over, you can build the app normally and it should not build the pre-built frameworks from sources. This can be verified by inspecting the build logs, but you can also examine the Pods targets in Xcode:
1) open xcworkspace in xcode
2) click Pods project from the file explorer (on the left)
3) find the "target" that corresponds to the library that should be prebuild on the "targets" list (i.e. "RNSVG" for svg or "react-native-webview" for webview – the target name depends on the podspec file name in the given library)
4) Go to "Build Phases" –> there should be no "compile sources" section there but instead a "copy xcframeworks" step

Future work:
1) We should remove the code that checks for artifacts in `/tmp/rnrepo`
2) We should use maven client to download artifacts such that we can take adventage of local maven cache. This can be done for example by using gradle (i.e. use gradle wrapper from android project when available)
3) We should figure out a way how to publish dSYMs for the prebuilt llibraries
4) Figure out a better way to distribute and add the plugin without requirung to publish in rubygems (not that I have anything agains it, but syncing versions and publishing to multiple different repositories is just hard to manage)
5) While testing I couldn't get release builds of app to pass, this is something we want to address in a follow up as this PR is already lengthy.

### DIFF
--- a/packages/builder/build-library-ios.ts
+++ b/packages/builder/build-library-ios.ts
@@ -4,7 +4,6 @@ import { join, basename } from 'path';
 import { sanitizePackageName } from '@rnrepo/config';
 import {
   type AllowedLicense,
-  getGithubBuildUrl,
   setupReactNativeProject,
 } from './build-utils';
 
@@ -55,7 +54,6 @@ if (buildConfig !== 'release' && buildConfig !== 'debug') {
 // Convert to Xcode configuration format (Release/Debug)
 const CONFIGURATION =
   buildConfig.charAt(0).toUpperCase() + buildConfig.slice(1);
-const GITHUB_BUILD_URL = getGithubBuildUrl();
 
 // Note: Bitcode was deprecated in Xcode 14 and removed entirely by Apple.
 // We build static frameworks instead, which provide better app size and performance.
--- a/packages/cocoapods-plugin/README.md
+++ b/packages/cocoapods-plugin/README.md
@@ -49,7 +49,7 @@ To test it on a project that lives outside of RNRepo monorepo structure, you can
 
 ```bash
 cd packages/cocoapods-plugin
-bun run install
+bun run gem-install
 ```
 
 This will automatically uninstall any previous version, build the gem, and install it.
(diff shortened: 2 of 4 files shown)


radoslawrolka (approved):
- Left some copilot comments for your consideration,
- added nitpicks fixes in [64d6b09](https://github.com/software-mansion/rnrepo/commit/64d6b09e532543a2cf0b2a33ec0cc6fae6fc3464)

PR #202, opened 2026-01-08
Rolkrado/gradle ios

radoslawrolka:
## 📝 Description

Added downloading from gradle if present

## 🎯 Type of Change

### DIFF
--- a/packages/cocoapods-plugin/build.gradle.kts
+++ b/packages/cocoapods-plugin/build.gradle.kts
@@ -1,48 +1,50 @@
-val defaultRnrepoUrl = System.getProperty("url") ?: "https://packages.rnrepo.org/releases"
+allprojects {
+    val defaultRnrepoUrl = System.getProperty("url") ?: "https://packages.rnrepo.org/releases"
 
-repositories {
-    mavenCentral()
-    google()
-    maven { url = uri(defaultRnrepoUrl) }
-}
+    repositories {
+        mavenCentral()
+        google()
+        maven { url = uri(defaultRnrepoUrl) }
+    }
 
-configurations.create("zipDownload") {
-    isCanBeConsumed = false
-    isCanBeResolved = true
-}
+    configurations.create("zipDownload") {
+        isCanBeConsumed = false
+        isCanBeResolved = true
+    }
 
-fun getArgument(property: String): String =
-    System.getProperty(property) 
-        ?: project.findProperty(property) as? String
-        ?: throw GradleException("Argument '$property' not provided. Use: gradle downloadArtifact -Dpackage='<pkg>' -Dversion='<version>' -DrnVersion='<rn-version>' -Dconfiguration='<release|debug>'")
+    fun getArgument(property: String): String =
+        System.getProperty(property) 
+            ?: project.findProperty(property) as? String
+            ?: throw GradleException("Argument '$property' not provided. Use: gradle downloadArtifact -Dpackage='<pkg>' -Dversion='<version>' -DrnVersion='<rn-version>' -Dconfiguration='<release|debug>'")
 
-// Task that accepts string arguments and downloads zip from Maven
-// Run: gradle downloadArtifact -Dpackage='<pkg>' -Dversion='<version>' -DrnVersion='<rn-version>' -Dconfiguration='<release|debug>' [-DworkletsVersion='<version>'] [-Durl='<rnrepo-url>']
-tasks.register("downloadArtifact") {
-    doLast {
-        val packageName = getArgument("package")
-        val version = getArgument("version")
-        val rnVersion = getArgument("rnVersion")
-        val configuration = getArgument("configuration")
-        val workletsVersion = System.getProperty("workletsVersion") ?: project.findProperty("workletsVersion") as? String
-        
-        val workletsSuffix = if (!workletsVersion.isNullOrEmpty()) "-worklets$workletsVersion" else ""
-        val fullNotation = "org.rnrepo.public:$packageName:$version:rn$rnVersion$workletsSuffix-$configuration"
-        
-        logger.info("[📦 RNRepo] Downloading $packageName version $version with RN version $rnVersion")
-        dependencies {
-            add("zipDownload", "$fullNotation.xcframework@zip")
-        }
-        
-        val resolvedFiles = configurations["zipDownload"].resolve()
-        if (resolvedFiles.isEmpty()) {
-            throw GradleException("No artifacts found for dependency: $fullNotation")
-        }
-        
-        resolvedFiles.forEach { file ->
-            logger.info("[📦 RNRepo] $fullNotation downloaded to: ${file.absolutePath}")
-            // Output the file path for external scripts to consume
-            println("DOWNLOADED_FILE:${file.absolutePath}")
+    // Task that accepts string arguments and downloads zip from Maven
+    // Run: gradle downloadArtifact -Dpackage='<pkg>' -Dversion='<version>' -DrnVersion='<rn-version>' -Dconfiguration='<release|debug>' [-DworkletsVersion='<version>'] [-Durl='<rnrepo-url>']
+    tasks.register("downloadArtifact") {
+        doLast {
+            val packageName = getArgument("package")
+            val version = getArgument("version")
+            val rnVersion = getArgument("rnVersion")
+            val configuration = getArgument("configuration")
+            val workletsVersion = System.getProperty("workletsVersion") ?: project.findProperty("workletsVersion") as? String
+            
+            val workletsSuffix = if (!workletsVersion.isNullOrEmpty()) "-worklets$workletsVersion" else ""
+            val fullNotation = "org.rnrepo.public:$packageName:$version:rn$rnVersion$workletsSuffix-$configuration"
+            
+            logger.info("[📦 RNRepo] Downloading $packageName version $version with RN version $rnVersion")
+            dependencies {
+                add("zipDownload", "$fullNotation.xcframework@zip")
+            }
+            
+            val resolvedFiles = configurations["zipDownload"].resolve()
+            if (resolvedFiles.isEmpty()) {
+                throw GradleException("No artifacts found for dependency: $fullNotation")
+            }
+            
+            resolvedFiles.forEach { file ->
+                logger.info("[📦 RNRepo] $fullNotation downloaded to: ${file.absolutePath}")
+                // Output the file path for external scripts to consume
+                println("DOWNLOADED_FILE:${file.absolutePath}")
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


packages/cocoapods-plugin/cocoapods-rnrepo.gemspec
```diff
@@ -13,13 +13,14 @@ Gem::Specification.new do |spec|
...
 
-  spec.files         = Dir['lib/**/*'] + %w{README.md}
+  spec.files         = Dir['lib/**/*'] + %w{README.md build.gradle.kts}
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ['lib']
 
   spec.add_dependency 'cocoapods', '>= 1.0.0'
   spec.add_dependency 'rubyzip', '~> 2.3'
+  spec.add_dependency 'openssl', '~> 4.0'
```
kmagiera: is this needed ? We dont use it anywhere now, you only use `open3` which I believe is something else and possibly a built-in package

packages/cocoapods-plugin/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
+repositories {
+    mavenCentral()
+    google()
+    //maven { url = uri("https://packages.rnrepo.org/releases/") }
+    maven { url = uri("https://repo.swmtest.xyz/releases/") }
```
kmagiera: can we make it so that the repository url is extacted from the arguments? I'd allow us to control the repo URL without changes across many different files / languages

packages/cocoapods-plugin/lib/cocoapods-rnrepo/downloader.rb
```diff
@@ -2,24 +2,101 @@
...
+
+      downloaded_path || Logger.log("Downloaded file not found in gradle cache at #{gradle_cache}")
+    end
+
+    # Download file via HTTP request
+    # Requires: artifact_spec hash with :sanitized_name, :version, :rn_version, :configuration, :destination
+    # Returns: destination path if successful, nil on failure
+    def self.download_via_http(artifact_spec)
+      # url = "https://packages.rnrepo.org/releases/org/rnrepo/public/#{artifact_spec[:sanitized_name]}/#{artifact_spec[:version]}/#{artifact_spec[:sanitized_name]}-#{artifact_spec[:version]}-rn#{artifact_spec[:rn_version]}-#{artifact_spec[:configuration]}.xcframework.zip"
+      url = "https://repo.swmtest.xyz/releases/org/rnrepo/public/#{artifact_spec[:sanitized_name]}/#{artifact_spec[:version]}/#{artifact_spec[:sanitized_name]}-#{artifact_spec[:version]}-rn#{artifact_spec[:rn_version]}-#{artifact_spec[:configuration]}.xcframework.zip"
```
kmagiera: would be good to have the repo URL in one place only. Now it is hardcoded here and also in the gradle file

packages/cocoapods-plugin/lib/cocoapods-rnrepo/downloader.rb
```diff
@@ -2,24 +2,101 @@
...
+        Logger.log "Error downloading via gradle: #{e.message}"
+        return nil
+      end
+    end
+
+    # Locate downloaded file in gradle cache
+    # Requires: artifact_spec hash with :sanitized_name, :version, :rn_version, :configuration
+    # Returns: path if found, nil if not found
+    def self.find_in_gradle_cache(spec)
+      path_parts = [Dir.home, '.gradle/caches/modules-2/files-2.1', 'org.rnrepo.public', spec[:sanitized_name], spec[:version]]
```
kmagiera: do we need to make some assumptions about the structure of the gradle cache? Was hoping that the gradle script could access the file location and return it, or perhaps we could pass the location where the artifacts should be unzipped and the gradle script would handle that

packages/cocoapods-plugin/lib/cocoapods-rnrepo/downloader.rb
```diff
@@ -2,24 +2,101 @@
...
+      false
+    end
+
+    # Download from local test files (development/testing only)
+    # Looks for files in: /tmp/rnrepo
+    # Requires: artifact_spec hash with :sanitized_name, :version, :rn_version, :configuration
+    # Returns: destination path if successful, nil if file not found
+    def self.download_from_local_test(artifact_spec)
+      local_test_dir = "/tmp/rnrepo"
+      filename = "#{artifact_spec[:sanitized_name]}-#{artifact_spec[:version]}-rn#{artifact_spec[:rn_version]}-#{artifact_spec[:configuration]}.xcframework.zip"
```
kmagiera: we don't handle the worklets version right now from what I can see. We should remember to add this in follow up

packages/cocoapods-plugin/lib/cocoapods-rnrepo/downloader.rb
```diff
@@ -2,24 +2,101 @@
...
 module CocoapodsRnrepo
   class Downloader
-    # Download file from URL
-    def self.download_file(url, destination)
-      # Check if file exists in local test directory first
-      filename = File.basename(url)
-      local_test_dir = '/tmp/rnrepo'
+
+    def self.gradle_installed?
+      _stdout, _stderr, status = Open3.capture3('gradle', '-v')
```
kmagiera: Wonder whether it is worth trying to use gradlew from android folder as the main gradle process as opposed to the gradle installed locally. The tradeoffs are:
1) if we use gradlew we stick to the gradle version they use for Android
2) if users don't run gradle frequently the wrapper will need to install itself which will take extra time.

We should also consider what happens on CIs

packages/cocoapods-plugin/cocoapods-rnrepo.gemspec
radoslawrolka: removed

packages/cocoapods-plugin/build.gradle.kts
radoslawrolka:
```
val default_rnrepo_url = System.getProperty("url") ?: "https://packages.rnrepo.org/releases/"

```

packages/cocoapods-plugin/lib/cocoapods-rnrepo/downloader.rb
radoslawrolka: added

packages/cocoapods-plugin/lib/cocoapods-rnrepo/downloader.rb
radoslawrolka: added class variable

packages/cocoapods-plugin/lib/cocoapods-rnrepo/downloader.rb
radoslawrolka: added gradle printout with path

packages/cocoapods-plugin/lib/cocoapods-rnrepo/downloader.rb
radoslawrolka: added gradlew from android with fallback
```
    def self.find_gradle_executable
      android_gradlew = File.join(Dir.pwd, '..', 'android', 'gradlew')
      @@gradle_executable = (File.executable?(android_gradlew) ? android_gradlew : 'gradle')
    end
```

packages/cocoapods-plugin/lib/cocoapods-rnrepo/downloader.rb
```diff
@@ -2,24 +2,128 @@
...
+require 'open3'
 
 module CocoapodsRnrepo
   class Downloader
-    # Download file from URL
-    def self.download_file(url, destination)
-      # Check if file exists in local test directory first
-      filename = File.basename(url)
-      local_test_dir = '/tmp/rnrepo'
+    @@repo_url = "https://packages.rnrepo.org/releases"
```
kmagiera, line 9: should we set this to `/snapshots` for the time being?

packages/cocoapods-plugin/lib/cocoapods-rnrepo/downloader.rb
```diff
@@ -2,24 +2,128 @@
...
+      Logger.log "Gradle check error: #{e.message}"
+      false
+    end
+
+    # Download from local test files (development/testing only)
+    # Looks for files in: /tmp/rnrepo
+    # Requires: artifact_spec hash with :sanitized_name, :version, :rn_version, :configuration, :destination, :worklets_version (optional)
+    # Returns: destination path if successful, nil if file not found
+    def self.download_from_local_test(artifact_spec)
+      local_test_dir = "/tmp/rnrepo"
```
kmagiera: I think it'd be cleaner to add some env variable to guard these checks such that it is possible for the packages to be loaded from some folder (i.e. `/tmp/rnrepo`) only if some debug env variable is set.

For example:
```
RNREPO_BUILDS_FOLDER=/tmp/rnrepo pod instal
```

packages/cocoapods-plugin/lib/cocoapods-rnrepo/downloader.rb
```diff
@@ -2,24 +2,128 @@
...
+      gem_root = File.expand_path('../../..', __FILE__)
+      gem_gradle_file = File.join(gem_root, 'build.gradle.kts')
+
+      unless File.exist?(gem_gradle_file)
+        Logger.log "build.gradle.kts not found at #{gem_gradle_file}"
+        return nil
+      end
+
+      # Write gradle file to ios directory to avoid permission issues
+      ios_gradle_file = File.join(rnrepo_dir, 'build.gradle.kts')
```
kmagiera: what are the permission issues here? I'd prefer that we avoid writing files to the source directories during install phase. Why can't we use the gradle file that we distribute as a part of this plugin? Maybe use the init script approach to load it?

packages/cocoapods-plugin/lib/cocoapods-rnrepo/downloader.rb
radoslawrolka: done

packages/cocoapods-plugin/lib/cocoapods-rnrepo/downloader.rb
radoslawrolka: I assume everyone who would work on the plugin will change it for whatever host/repo they have their artifacts hosted on.

packages/cocoapods-plugin/lib/cocoapods-rnrepo/downloader.rb
radoslawrolka: Fixed with initscript


`Build.gradle.kts` from `cocoapods-plugin/` is packed into `.gem ` which is then installed on the user machine with path similar to below:

```
/Users/<username>/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/cocoapods-rnrepo-0.1.0/build.gradle.kts
```

Inside that directory we do not have write permissions for automatic gradle cache files, so it had (now init script) to be run elsewhere

PR #203, opened 2026-01-09
Rolkrado/add pickfirsts

radoslawrolka:
## 📝 Description

[FIX] Add pickfirsts for "react-native-nitro-modules"="libNitroModule
[TROUBLESHOOTING] Add empty repos solution

## 🎯 Type of Change

Please delete options that are not relevant:

- [x] 📚 Documentation update

### DIFF
--- a/packages/gradle-client/rnrepo-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
+++ b/packages/gradle-client/rnrepo-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
@@ -210,8 +210,8 @@ class PrebuildsPlugin : Plugin<Project> {
         }
 
         prebuiltSoProvidersPackageAndLibName.forEach { (prebuiltSoProvidersPackageName, nativeLibName) ->
-            val isprebuiltSoProvidersPackageSupported = supportedPackages.any { it.name == prebuiltSoProvidersPackageName }
-            if (!isprebuiltSoProvidersPackageSupported) {
+            val isPrebuiltSoProvidersPackageSupported = supportedPackages.any { it.name == prebuiltSoProvidersPackageName }
+            if (!isPrebuiltSoProvidersPackageSupported) {
                 logger.info("Provider package '$prebuiltSoProvidersPackageName' is not supported, skipping pickFirsts configuration.")
                 return@forEach
             }


TROUBLESHOOTING.md
```diff
@@ -106,3 +106,49 @@ These messages indicate that the RNRepo plugin has detected an inconsistency and
...
+ - /path/to/node_modules/react-native-some-package/android/build/intermediates/library_jni/debug/copyDebugJniLibsProjectOnly/jni/arm64-v8a/libName.so
+ - /path/to/.gradle/caches/.../react-native-some-package-0.0.0-rn0.81.4/jni/arm64-v8a/libName.so
+```
+
+This error occurs when:
+1. A library built from sources includes native code from a provider library
+2. The Gradle build system encounters duplicate `.so` files in different locations during the merge phase
+
+#### Automatic Solution
+The RNRepo plugin automatically detects when both a provider library (e.g., `react-native-worklets`) and its consumer library (e.g., `react-native-reanimated`) are supported as prebuilts. When this happens, it adds `pickFirsts` configuration for all native library files to resolve the conflict gracefully.
```
WoLewicki: Do both of those need to be prebuilts? I guess one is enough? Or even none of those, we still apply the logic yes?

TROUBLESHOOTING.md
```diff
@@ -106,3 +106,49 @@ These messages indicate that the RNRepo plugin has detected an inconsistency and
...
+```
+
+Or in info logs:
+```
+[📦 RNRepo] HTTP RNRepo repositories to check: <no repositories listed>
+```
+
+This typically means that:
+1. Your project has packages that could be prebuilt, but none are being recognized as supported (i.e., unsupported packages/versions, or all are in the deny list)
+2. The plugin cannot find any configured Maven repositories pointing to the RNRepo registry
```
WoLewicki, line 150: Could we handle this situation in plugin by checking the list of maven repositories and if ours is not found, posting a warning or something?

packages/gradle-client/rnrepo-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
```diff
@@ -202,6 +190,39 @@ class PrebuildsPlugin : Plugin<Project> {
...
+            )
+        val architectures = listOf("arm64-v8a", "armeabi-v7a", "x86", "x86_64")
+        val androidExtension = project.extensions.getByName("android") as? BaseExtension
+        if (androidExtension == null) {
+            logger.warn("Android extension not found in project ${project.name}, cannot configure pickFirsts.")
+            return
+        }
+
+        consumedPackageAndLibName.forEach { (consumedPackageName, nativeLibName) ->
+            val isConsumedPackageSupported = supportedPackages.any { it.name == consumedPackageName }
```
WoLewicki: Is consumed the proper keyword? We are interested in the libraries that are prebuilt and provide `.so` files that can be used by other libraries. So maybe we should call them prebuiltSoProviders or something like that?

TROUBLESHOOTING.md
radoslawrolka: changed ```
The RNRepo plugin automatically detects when a provider library (e.g., `react-native-worklets`) or its consumer library (e.g., `react-native-reanimated`) are present. When this happens, it adds `pickFirsts` configuration for all native library files to resolve the conflict gracefully.
```

File: TROUBLESHOOTING.md (hunk 2 above)
Comment (radoslawrolka) [reply TROUBLESHOOTING.md:150]: added link, will work when this pr will be merged in main
```diff
        if (httpRepositories.isEmpty()) {
+            logger.warn("No RNRepo maven repository found to check for packages. Check your gradle configuration (https://github.com/software-mansion/rnrepo/blob/main/TROUBLESHOOTING.md#no-supported-packages-found-or-empty-repository-list).")
            return false
        }
```

packages/gradle-client/rnrepo-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
radoslawrolka: changed to `prebuiltSoProviders`

TROUBLESHOOTING.md
WoLewicki: Why do we need consumer libraries here? I guess it is only important if there is a library that exposes the `.so` in the project, and in such case we add the `pickFirsts` ? They prolly don't even need to be prebuilt since it won't ever destroy any behavior right? Or am I missing something?

radoslawrolka: Right, removed the part about the consumer

packages/gradle-client/rnrepo-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
```diff
@@ -202,6 +190,39 @@ class PrebuildsPlugin : Plugin<Project> {
...
+            )
+        val architectures = listOf("arm64-v8a", "armeabi-v7a", "x86", "x86_64")
+        val androidExtension = project.extensions.getByName("android") as? BaseExtension
+        if (androidExtension == null) {
+            logger.warn("Android extension not found in project ${project.name}, cannot configure pickFirsts.")
+            return
+        }
+
+        prebuiltSoProvidersPackageAndLibName.forEach { (prebuiltSoProvidersPackageName, nativeLibName) ->
+            val isprebuiltSoProvidersPackageSupported = supportedPackages.any { it.name == prebuiltSoProvidersPackageName }
```
WoLewicki:
```suggestion
            val isPrebuiltSoProvidersPackageSupported = supportedPackages.any { it.name == prebuiltSoProvidersPackageName }
```

Same below

PR #205, opened 2026-01-09
[iOS] Add working ci pipeline for ios xcframeworks

radoslawrolka:
## 📝 Description

- Working ci pipeline building and publishing artifacts (release and debug)
- Using prebuilt RN instead of building from sources 
- building just pod instead of whole app

Diff will shrink, when kmagiera/ios will be merged to main
## 🎯 Type of Change

### DIFF
--- a/packages/publisher/publish-library-ios.ts
+++ b/packages/publisher/publish-library-ios.ts
@@ -112,7 +112,7 @@ async function main() {
     // Find the XCFramework zip files for both Release and Debug builds
     for (const buildConfig of ['release', 'debug']) {
       console.log(`\n🔍 Searching for ${buildConfig} build...`);
-      const xcframeworkZip = `${sanitizedLibraryName}-${libraryVersion}-rn${reactNativeVersion}${workletsVersion ? `-worklets${workletsVersion}` : ''}-${buildConfig}.xcframework.zip`;
+      const xcframeworkZip = `${sanitizedLibraryName}-${libraryVersion}-rn${reactNativeVersion}-${buildConfig}.xcframework.zip`;
       const xcframeworkPath = join(artifactDir, xcframeworkZip);
 
       if (!existsSync(xcframeworkPath)) {


.github/workflows/publish-library-ios.yml
```diff
@@ -0,0 +1,103 @@
...
+    workflows: [Build iOS Library]
+    types: [completed]
+
+jobs:
+  handle-build-completion:
+    runs-on: rnrepo-ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+    if: github.event.workflow_run.conclusion != 'success' && !contains(github.event.workflow_run.display_title, 'snapshot')
```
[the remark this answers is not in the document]
radoslawrolka: `publish-library` job is for successful jobs, `handle-build-completion` job is for all other results

.github/workflows/build-library-ios.yml
```diff
@@ -29,7 +29,8 @@ on:
 
 jobs:
   build-library-ios:
-    runs-on: rnrepo-builder-macos-14
+    runs-on: macos-latest
+    # runs-on: rnrepo-builder-macos-14
```
kmagiera: github doesn't allow for creating custom macos runners at the moment (unless self-hosted), so we won't be using the custom group name any time soon. Can delete this

packages/builder/build-library-ios.ts
```diff
@@ -171,103 +155,105 @@ async function buildFramework(appDir: string, _license: AllowedLicense) {
...
-    console.log('🔍 Locating built framework...');
-
-    // CocoaPods may sanitize pod names to valid C identifiers (e.g., dashes → underscores)
-    // So we glob for *.framework instead of assuming the sanitization logic
-    const podBuildDir = join(
-      buildDir,
-      `${CONFIGURATION}-iphonesimulator`,
-      podName
-    );
+  for (const CONFIGURATION of buildConfigs) {
```
kmagiera: use a lowercase letters for variable name

packages/publisher/publish-library-ios.ts
```diff
@@ -0,0 +1,194 @@
...
+        `Framework artifacts directory not found at ${artifactDir}`
+      );
+    }
+
+    // List available files to find the xcframework zip
+    const files = readdirSync(artifactDir);
+    for (const buildConfig of ['release', 'debug']) {
+      console.log(`\n🔍 Searching for ${buildConfig} build...`);
+      const xcframeworkZip = files.find((f) =>
+        f.includes(`${sanitizedLibraryName}-${libraryVersion}`) &&
```
kmagiera: why do we find files like this. Seems like we should be able to tell the exact name of the file. It would be safer to expect a concrete name such that we don't leave any room for abusing this code path in case some malicious library would affect the build script and what lands in the artifacts

packages/publisher/publish-library-ios.ts
```diff
@@ -0,0 +1,194 @@
...
+          -DgeneratePom=false \
+          -Durl=${MAVEN_REPOSITORY_URL}`;
+      console.log('✓ XCFramework deployed successfully');
+
+      console.log(
+        `✅ Published library ${libraryName}@${libraryVersion} (${buildConfig}) to remote Maven repository`
+      );
+    }
+    // Update Supabase status to 'completed' after publish is fully complete
+    if (isSnapshotRun) {
```
kmagiera, line 151: Not sure why we don't do this for Android but shouldn't snapshot runs use snapshot repository URL?

.github/workflows/build-library-ios.yml
radoslawrolka: done

packages/builder/build-library-ios.ts
radoslawrolka: done

packages/publisher/publish-library-ios.ts
radoslawrolka: It is set in yml file:
- https://github.com/software-mansion/rnrepo/blob/main/.github/workflows/publish-library-android.yml#L98
- https://github.com/software-mansion/rnrepo/blob/main/.github/workflows/publish-library-ios.yml#L98
```
      - name: 📤 Publish library
        run: bun run packages/publisher/publish-library-ios.ts -- "${{ github.event.workflow_run.id }}"
        env:
          ...
          MAVEN_REPOSITORY_URL: ${{ format('{0}{1}', secrets.RNREPO_SWM_REPOSITORY_URL_BASE, contains(github.event.workflow_run.display_title, 'snapshot') && '/snapshots' || '/releases') }}
          ...
```

packages/publisher/publish-library-ios.ts
radoslawrolka: done

packages/publisher/update-build-status.ts
```diff
@@ -68,18 +68,19 @@ async function main() {
       throw new Error('Could not get build workflow run name');
     }
 
-    // Format: "Build for Android {library_name}@{library_version} RN@{react_native_version}( with worklets@{worklets_version})?"
-    const match = buildRunName.match(/Build for Android (.+?)@(.+?) RN@(.+?)( with worklets@(.+?))?$/);
+    // Format: "Build for {platform} {library_name}@{library_version} RN@{react_native_version}( with worklets@{worklets_version})?"
+    const match = buildRunName.match(/Build for (Android|iOS) (.+?)@(.+?) RN@(.+?)( with worklets@(.+?))?$/);
```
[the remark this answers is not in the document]
radoslawrolka: snapshots are handled in yml file, so updates_status is not even executed for snapshots builds

PR #206, opened 2026-01-13
feat: codegen in android packages

WoLewicki:
## 📝 Description

PR adding codegen artifacts (static libs and headers) as parts of library `.aar` and implementing:
- packing all of it to `.aar`
- unpacking them in client project
- linking them instead of building them from source by:
	- removing library occurences from `Android-autolinking.cmake` (https://github.com/software-mansion/rnrepo/pull/206/files#diff-8d1431ccdc7a315b68ce17002287169c0463238f411af18e09163e67e34a0334R924-R938)
	- adding `CMakeLists.txt` wrapper around the main app's one in which we manually link those libs (https://github.com/software-mansion/rnrepo/pull/206/files#diff-8d1431ccdc7a315b68ce17002287169c0463238f411af18e09163e67e34a0334R999) which is consumed [here](https://github.com/facebook/react-native/blob/f9346fb23432dab8a58d51225131805b105adcc8/packages/gradle-plugin/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/utils/NdkConfiguratorUtils.kt#L29-L35)

It is done to speed up the builds by removing the necessity of building codegen each time you build an app since we believe it should be library's responsibility to provide those artifacts.

### Patching

patch is successfully applied for all react-native@>=0.81.4 versions.
```
 PATCH COMPATIBILITY RESULTS
RN VERSION    RESULT
──────────────────────────────────
0.77.3        ✗ patch mismatch
0.78.3        ✗ patch mismatch
0.79.7        ✗ patch mismatch
0.80.0        ✗ patch mismatch
0.80.1        ✗ patch mismatch
0.80.2        ✗ patch mismatch
0.80.3        ✗ patch mismatch
0.81.0        ✓
0.81.1        ✓
0.81.4        ✓
0.81.5        ✓
0.81.6        ✓
0.82.0        ✓
0.82.1        ✓
0.83.0        ✓
0.83.1        ✓
0.83.2        ✓
0.83.4        ✓
0.84.0        ✓
```


packages/builder/patches/@react-native+codegen+0.81.4.patch
WoLewicki, line 1: The patch is for rn 0.81.4 but as long as the code in those files won't change in rn repo, it should work with other versions too.

packages/publisher/publish-library-android.ts
```diff
@@ -124,7 +124,7 @@ async function main() {
     const pomFile = join(artifactsBasePath, `${baseFileName}.pom`);
     const classifier = `rn${reactNativeVersion}${
       workletsVersion ? `-worklets${workletsVersion}` : ''
-    }`;
+    }-codegen`;
```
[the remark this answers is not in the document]
radoslawrolka: done

packages/build-tools/gradle-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
kmagiera: is it ok that we don't consinder `-codegen` suffix here?

packages/builder/build-utils.ts
```diff
@@ -335,6 +341,12 @@ export async function setupReactNativeProject(
   console.log('📦 Installing all dependencies...');
   await $`npm install`.cwd(appDir).quiet();
 
+  // check if there is patches folder to apply
+  if (existsSync(join(appDir, 'patches'))) {
+    console.log('📦 Applying patches...');
+    await $`npx patch-package`.cwd(appDir);
```
kmagiera: would prefer to not use `npx` here but use some fixed version of the tool instead

kmagiera: will this not fail if there are no patches?

packages/builder/patches/@react-native+codegen+0.81.4.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@react-native/codegen/lib/generators/modules/GenerateModuleJniH.js b/node_modules/@react-native/codegen/lib/generators/modules/GenerateModuleJniH.js
```
kmagiera, line 16: why do we need this patch? This needs to be explained at least in the PR description

Why we only need this for the specific 0.81.4 version and not for others? Do we even need to support codegen prebuilds for this relatively old version of RN?

packages/builder/gradle_init_scripts/codegen-build.gradle
```diff
@@ -0,0 +1,32 @@
...
+        // we want to cut it in the main target since it builds the codegen artifacts
+        def isAndroidApp = project.plugins.hasPlugin("com.android.application")
+        if (isAndroidApp) {
+            project.android {
+                buildTypes {
+                    debug {
+                        externalNativeBuild {
+                            cmake {
+                                def currentArgs = arguments ?: []
+                                // currentArgs += "-DCMAKE_BUILD_TYPE=Debug"
```
kmagiera: delete this, instead add comment explaining why we add flags and what they are for

packages/builder/gradle_init_scripts/codegen-build.gradle
```diff
@@ -0,0 +1,32 @@
...
+                                currentArgs += "-DCMAKE_CXX_FLAGS=-g1"
+                                arguments = currentArgs
+                            }
+                        }
+                    }
+                    release {
+                        externalNativeBuild {
+                            cmake {
+                                def currentArgs = arguments ?: []
+                                // currentArgs += "-DCMAKE_BUILD_TYPE=Release"
```
kmagiera: same here

packages/publisher/publish-library-android.ts
```diff
@@ -122,13 +122,25 @@ async function main() {
...
-    const aarFile = join(
+    let aarFile = join(
       artifactsBasePath,
       `${baseFileName}-${classifier}.aar`
     );
+    if (!existsSync(aarFile)) {
+      classifier += '-codegen';
+      aarFile = join(
+        artifactsBasePath,
+        `${baseFileName}-${classifier}.aar`
```
kmagiera: this is a very weird code construct. Why we're changing classifier to `let` only to append "-codegen" suffix to get the aarFile name.

I think it'd be cleaer to just make separate `aatFile` and `aarFileWithCodegen`, then have a cleaner check that at least one should exist.

packages/builder/build-library-android.ts
```diff
@@ -105,41 +188,112 @@ async function buildAAR(appDir: string, license: AllowedLicense) {
...
-    const aarPath = join(
-      mavenLocalLibraryLocationPath,
-      `${gradleProjectName}-${libraryVersion}-${classifier}.aar`
-    );
-    if (!existsSync(aarPath)) {
-      throw new Error(`AAR file not found at ${aarPath}`);
+    // If library has codegen, build codegen version as well
+    if (!hasCodegenConfig) {
+      console.log('ℹ️ No codegen configuration found, building just standard AAR version');
+      const args = [
```
kmagiera: maybe extract this and the other build calls into separate methods. These `if` blocks here are very long and make things harder to read/follow

packages/builder/build-library-android.ts
```diff
@@ -105,41 +188,112 @@ async function buildAAR(appDir: string, license: AllowedLicense) {
...
-    }
-    const aarPath = join(
-      mavenLocalLibraryLocationPath,
-      `${gradleProjectName}-${libraryVersion}-${classifier}.aar`
-    );
-    if (!existsSync(aarPath)) {
-      throw new Error(`AAR file not found at ${aarPath}`);
+    // If library has codegen, build codegen version as well
+    if (!hasCodegenConfig) {
+      console.log('ℹ️ No codegen configuration found, building just standard AAR version');
```
kmagiera: remove "just"

packages/builder/build-library-android.ts
```diff
@@ -105,41 +188,112 @@ async function buildAAR(appDir: string, license: AllowedLicense) {
...
+        `-PrnrepoArtifactId=${gradleProjectName}`,
+        `-PrnrepoPublishVersion=${libraryVersion}`,
+        `-PrnrepoClassifier=${classifier}`,
+        `-PrnrepoCpuInfo=${getCpuInfo()}`,
+        `-PrnrepoBuildUrl=${GITHUB_BUILD_URL}`,
+        `-PrnrepoLicenseName=${license}`,
+        `-PrnrepoLicenseUrl=https://opensource.org/license/${license}`,
+      ];
+      await $`./gradlew ${args}`.cwd(androidPath);
+      // verify that the .pom and .aar files are present aftre the publish command completes
```
kmagiera:
```suggestion
      // verify that the .pom and .aar files are present after the publish command completes
```

packages/builder/build-library-android.ts
```diff
@@ -105,41 +188,112 @@ async function buildAAR(appDir: string, license: AllowedLicense) {
...
+        '--init-script', codegenBuildGradleScriptPath,
+        ...(postinstallGradleScriptPath
+          ? ['--init-script', postinstallGradleScriptPath]
+          : []),
+        `-PrnrepoCodegenName=${packageJson.codegenConfig.name}`,
+      ];
+      await $`./gradlew ${assembleReleaseArgs}`.cwd(androidPath);
+
+      console.log('📦 Publishing codegen version...');
+      const args = [
```
kmagiera: isn't some part of the build/publish logic duplicated here and in the previous blocks?

radoslawrolka:
After building artifacts are the same, performed tests to check for regressions, tests from #368 are still valid.
```
$ shasum -a 256 ~/.m2/repository/org/rnrepo/public/react-native-safe-area-context/5.8.0/react-native-safe-area-context-5.8.0-rn0.85.1-codegen.aar
shasum -a 256 ~/Downloads/RNREPO_STASH/CODEGEN_ARTIFACTS_FOR_PATCH_TEST/react-native-safe-area-context/5.8.0/react-native-safe-area-context-5.8.0-rn0.85.1-codegen.aar
cf84238baed548915cc72a3493358aad575dc5411e70b5215596290d3601e624  ~/.m2/repository/org/rnrepo/public/react-native-safe-area-context/5.8.0/react-native-safe-area-context-5.8.0-rn0.85.1-codegen.aar
cf84238baed548915cc72a3493358aad575dc5411e70b5215596290d3601e624  ~/Downloads/RNREPO_STASH/CODEGEN_ARTIFACTS_FOR_PATCH_TEST/react-native-safe-area-context/5.8.0/react-native-safe-area-context-5.8.0-rn0.85.1-codegen.aar
---
$ shasum -a 256 ~/.m2/repository/org/rnrepo/public/react-native-screens/4.25.1/react-native-screens-4.25.1-rn0.85.1-codegen.aar
shasum -a 256 ~/Downloads/RNREPO_STASH/CODEGEN_ARTIFACTS_FOR_PATCH_TEST/react-native-screens/4.25.1/react-native-screens-4.25.1-rn0.85.1-codegen.aar
d5f55293fe416068fff502ddf8ce04393da182ecdd024f071912ff3751ef8abf  ~/.m2/repository/org/rnrepo/public/react-native-screens/4.25.1/react-native-screens-4.25.1-rn0.85.1-codegen.aar
d5f55293fe416068fff502ddf8ce04393da182ecdd024f071912ff3751ef8abf  ~/Downloads/RNREPO_STASH/CODEGEN_ARTIFACTS_FOR_PATCH_TEST/react-native-screens/4.25.1/react-native-screens-4.25.1-rn0.85.1-codegen.aar
```

kmagiera (approved):
Looks good now 👍

radoslawrolka:
Latest changes works in rn-app and expo app 
```
    "react-native-gesture-handler": 2.31.2
    "react-native-reanimated": "4.3.1",
    "react-native-safe-area-context": "5.7.0",
    "react-native-screens": "4.25.2",
    "react-native-worklets": "0.8.3"
```
and checks in this pr assured no reqression was introduced
- https://github.com/software-mansion/rnrepo/actions/runs/28448028426
- https://github.com/software-mansion/rnrepo/actions/runs/28448028415

PR #213, opened 2026-01-15
nitpick pickfirst refactor

radoslawrolka:
## 📝 Description

Always add pickfirst if package is in project, so that prebuilds resolution works correctly

## 🎯 Type of Change


kmagiera (commented):
Why is this needed? I understand we need pickfirst because reanimated aars have libworklets. But in that instance the solutions should be to not distribute libworklets with reanimated. The PR description should mention the motivation for this change

Issue #216, opened 2026-01-15
[BUG] Failed on production build


ccorneliusHsv:
### 📋 Description

Tried adding this to our project. It built fine for dev and a test build with Firebase Distribution. But when I went to build a production version for the store, it failed. I'm not sure I have enough information to reproduce the issue. Feel free to close this issue if there not enough information to be helpful. 

The error was:

> Task :react-native-reanimated:preReleaseBuild
> Task :react-native-reanimated:configureCMakeRelWithDebInfo[arm64-v8a]

> Task :react-native-reanimated:buildCMakeRelWithDebInfo[arm64-v8a][reanimated] FAILED
C/C++: ninja: error: '/home/runner/work/react-native-app/react-native-app/node_modules/react-native-worklets/android/build/intermediates/cmake/release/obj/arm64-v8a/libworklets.so', needed by '../../../../build/intermediates/cxx/RelWithDebInfo/3z63p216/obj/arm64-v8a/libreanimated.so', missing and no known rule to make it

>FAILURE: Build failed with an exception.
>ASM Instrumentation process wasn't able to resolve some classes, this means that

>* What went wrong:
Execution failed for task ':react-native-reanimated:buildCMakeRelWithDebInfo[arm64-v8a][reanimated]'.
> com.android.ide.common.process.ProcessException: ninja: Entering directory `/home/runner/work/react-native-app/react-native-app/node_modules/react-native-reanimated/android/.cxx/RelWithDebInfo/3z63p216/arm64-v8a'
  [0/2] Re-checking globbed directories...
  
>  C++ build system [build] failed while executing:
      /usr/local/lib/android/sdk/cmake/3.22.1/bin/ninja \
        -C \
        /home/runner/work/react-native-app/react-native-app/node_modules/react-native-reanimated/android/.cxx/RelWithDebInfo/3z63p216/arm64-v8a \
        reanimated
    from /home/runner/work/react-native-app/react-native-app/node_modules/react-native-reanimated/android
  ninja: error: '/home/runner/work/react-native-app/react-native-app/node_modules/react-native-worklets/android/build/intermediates/cmake/release/obj/arm64-v8a/libworklets.so', needed by '../../../../build/intermediates/cxx/RelWithDebInfo/3z63p216/obj/arm64-v8a/libreanimated.so', missing and no known rule to make it

I looked to see if the folder it was looking for was there "node_modules/react-native-worklets/android/build/intermediates/cmake/release/" and only found a debug folder. 

[...]

PR #217, opened 2026-01-19
[GRADLE PLUGIN] App prefix warning

radoslawrolka:
## 📝 Description

Adds warning if plugin was launched from non app 

## 🎯 Type of Change

## 🧪 Testing

Zrzut ekranu 2026-01-19 o 10 23 41

## 📋 Additional Notes

Follows #216


kmagiera (commented):
Isn't there a better way to solve this? What are the issues with launching the command without `:app:` specifier? If this makes the build process fail completely, we shouldn't run the plugin. Or maybe we can limit the plugin to only scanning tasks that are within the `:app:` ??

PR #218, opened 2026-01-19
[PUBLISHER] Update status on failed publishing

radoslawrolka:
## 📝 Description

 Update status on failed publishing, so no build entries in db will be left hanging on 'scheduled'

## 🎯 Type of Change

### DIFF
--- a/packages/publisher/publish-library-android.ts
+++ b/packages/publisher/publish-library-android.ts
@@ -157,25 +157,26 @@ async function main() {
     // Sign and deploy AAR using gpg:sign-and-deploy-file (signs and deploys in one step)
     // The task uses MAVEN_GPG_KEY and MAVEN_GPG_PASSPHRASE environment variables to sign the artifact
     let publishStatus = 'failed' as 'failed' | 'completed';
-    try {
-      await $`mvn gpg:sign-and-deploy-file \
-          -Dfile=${aarFile} \
-          -DgroupId=org.rnrepo.public \
-          -DartifactId=${mavenLibraryName} \
-          -Dversion=${libraryVersion} \
-          -Dpackaging=aar \
-          -Dclassifier=${classifier} \
-          -DgeneratePom=false \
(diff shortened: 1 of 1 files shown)


packages/publisher/publish-library-android.ts
```diff
@@ -156,21 +156,28 @@ async function main() {
...
-        -DrepositoryId=RNRepo \
-        -Durl=${MAVEN_REPOSITORY_URL}`;
-    console.log('✓ AAR signed and deployed successfully');
-
-    console.log(
-      `✅ Published library ${libraryName}@${libraryVersion} to remote Maven repository`
-    );
+    let publishStatus = 'failed' as 'failed' | 'completed';
+    try {
+      await $`mvn gpg:sign-and-deploy-file \
```
kmagiera: alternatively we could use `nothrow()` and read the result of the shell call. IMO it'd be cleaner but not going to block this on a minor detail.

packages/publisher/publish-library-android.ts
```diff
@@ -156,21 +156,28 @@ async function main() {
...
+          -DgeneratePom=false \
+          -DrepositoryId=RNRepo \
+          -Durl=${MAVEN_REPOSITORY_URL}`;
+      console.log('✓ AAR signed and deployed successfully');
+
+      console.log(
+        `✅ Published library ${libraryName}@${libraryVersion} to remote Maven repository`
+      );
+      publishStatus = 'completed';
+    } catch (error) {
```
kmagiera: do we need that final catch in this case?

packages/publisher/publish-library-android.ts
radoslawrolka: changed to nothrow

packages/publisher/publish-library-android.ts
radoslawrolka: added for logging the error

PR #219, opened 2026-01-19
Move cocoapods files and adapt for ruby scripting

radoslawrolka:
## 📝 Description

Adapt ruby scripts for npm package that will works when launched from podfile:
- removed all gem references - unused in current usecase 
- moved files after removing the gem plugin entrypoint

## 🎯 Type of Change

### DIFF
--- a/packages/cocoapods-plugin/lib/downloader.rb
+++ b/packages/cocoapods-plugin/lib/downloader.rb
@@ -2,7 +2,6 @@ require 'net/http'
 require 'uri'
 require 'fileutils'
 require 'zip'
-require 'open3'
 
 module CocoapodsRnrepo
   class Downloader


packages/cocoapods-plugin/lib/plugin.rb
```diff
@@ -331,19 +351,26 @@ class << self
 
       CocoapodsRnrepo::Logger.log ""
     end
+    
+    # Register post_install hook inside the monkey patch
+    unless @rnrepo_post_install_registered
+      Pod::HooksManager.register('cocoapods-rnrepo', :post_install) do |installer_context|
```
kmagiera: I thought we no longer use the plugin approach and hence calling into HooksManager shouldn't be needed? What is the reason for this change?

packages/cocoapods-plugin/lib/downloader.rb
```diff
@@ -150,8 +81,7 @@ def self.download_file(artifact_spec)
       result = download_from_local_test(artifact_spec)
```
kmagiera: we should mark this to be cleaned up later or even delete this as part of this PR if there's not much use for this method anymore

radoslawrolka: removed

PR #236, opened 2026-01-28
Gradle - fix uncatched dependencies

radoslawrolka:
## 📝 Description

Improved the logic in `PrebuildsPlugin.kt` to more accurately determine when a package with C++ dependencies should fall back to source builds, by checking the presence and support status of dependent packages.

Not all packages were fallbacking to sources when provider pacakge was built from sources

When any firebase/* was not supported and some firebase/* were supported with aar, plugin did not pushed all supported firebase/* libs to build from sources (image displays the resolved packages before the fix and after the fix)
  
Also added 4 new firebase libs

## 🎯 Type of Change

## 🧪 Testing

- builded with firebase packages - diff between 0.2.2 vs patched plugin

Zrzut ekranu 2026-01-27 o 20 27 57

### DIFF
--- a/packages/build-tools/gradle-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
+++ b/packages/build-tools/gradle-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
@@ -692,7 +692,7 @@ class PrebuildsPlugin : Plugin<Project> {
                 projectPackages
                     .filter { dependentPackagePatternRegex.matches(it.name) && it.name != packageItem.name }
                     .filterNot { dep -> supportedPackages.any { it.name == dep.name } }
-            if (!allDependentPackageAreSupported.isEmpty()) {
+            if (allDependentPackageAreSupported.isNotEmpty()) {
                 unavailablePackages.add(packageItem)
                 val packagesToCheckRegex = PACKAGES_WITH_CPP[packageItem.name]?.map { it.toRegex() } ?: emptyList()
                 val packagesToRemove =


packages/build-tools/gradle-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
```diff
@@ -812,13 +812,13 @@ class PrebuildsPlugin : Plugin<Project> {
         dependentList.parallelStream().forEach { packageItem ->
             logger.info("Handling ${packageItem.npmName} package after others.")
             val elseClosure: () -> Unit = {
-                if (getBuildType(project) == "release") {
+                if (getBuildType(project) == "release" && packageItem.name != "react-native-firebase_app") {
```
kmagiera: I don't understand why we special case this specific name in a radon code place like here. This at least deserves an inline comment to explain

radoslawrolka: Added com: Firebase packages fails even in release builds, so we have to validate all Firebase packages regardless of build type.

packages/build-tools/gradle-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
```diff
@@ -812,13 +812,14 @@ class PrebuildsPlugin : Plugin<Project> {
         dependentList.parallelStream().forEach { packageItem ->
             logger.info("Handling ${packageItem.npmName} package after others.")
             val elseClosure: () -> Unit = {
-                if (getBuildType(project) == "release") {
+                // Firebase packages fails even in release builds, so we have to validate all Firebase packages regardless of build type.
+                if (getBuildType(project) == "release" && packageItem.name != "react-native-firebase_app") {
```
kmagiera: lets remove this conditions. It looks like it only applies to reanimated + worklets now. There's no guarantee other packages will work in this setup

PR #237, opened 2026-01-28
create @rnrepo/build-tools npm package

radoslawrolka:
## 📝 Description

- moved all cocapods & gradle plugins as subdirs into packages/build-tools
- adjusted CI
- updated docs

## 🎯 Type of Change

### DIFF
--- a/packages/build-tools/package.json
+++ b/packages/build-tools/package.json
@@ -6,8 +6,7 @@
   "scripts": {
     "build": "./gradle-plugin/gradlew build --no-daemon -p gradle-plugin",
     "verify": "bun run ./scripts/verify-plugin-version-before-publish.ts",
-    "prepublishOnly": "npm run build",
-    "prepare": "npm run verify"
+    "prepublishOnly": "npm run build && npm run verify"
   },
   "author": "Software Mansion",
   "license": "ISC",


README.md
```diff
@@ -66,13 +66,9 @@ For standard React Native setups or when using Expo but managing your android fo
...
    ```diff
    buildscript {
-     repositories {
-       ...
-   +   maven { url "https://packages.rnrepo.org/releases" }
-     }
      dependencies {
        ...
-   +   classpath("org.rnrepo.tools:prebuilds-plugin:0.2.2")
+   +   classpath fileTree(dir: "../node_modules/@rnrepo/prebuilds-plugin/gradle-plugin/build/libs", include: ["prebuilds-plugin-*.jar"])
```
kmagiera: as discussed offline, we should rely on the node resolve mechanism here instead

.github/workflows/publish-prebuilds-plugin.yml
```diff
@@ -0,0 +1,66 @@
...
+
+      - name: 🏗 Update npm
+        run: npm install -g npm@latest
+      
+      - name: 🔨 Build RNRepo Gradle Plugin for Android
+        run: ./gradlew build --no-daemon
+        working-directory: ./packages/prebuilds-plugin/gradle-plugin
+
+      - name: ✅ Verify JAR version matches build.gradle
+        run: bun run ./.github/workflows/scripts/verify-plugin-version-before-publish.ts
```
kmagiera: I wonder if we really should have this step here. When publishing we always build the plugin, so perhaps the build command should also be makins sure that the versions are in sync. One way would be for the build script to take version as argument.

README.md
radoslawrolka: done

.github/workflows/publish-prebuilds-plugin.yml
radoslawrolka: removed building & verify, that are executed in npm publish

PR #241, opened 2026-02-05
Fix iOS Release builds

radoslawrolka:
## 📝 Description
Refactor RNREPO build phase script handling and improve symlink creation for framework configurations
 - unused variables will be removed in lint-ruby PR
 - 

## 🎯 Type of Change

### DIFF
--- a/packages/build-tools/cocoapods-plugin/lib/plugin.rb
+++ b/packages/build-tools/cocoapods-plugin/lib/plugin.rb
@@ -226,15 +226,13 @@ module Pod
         # Find the xcframework (name may differ from pod name due to sanitization)
         cache_dir = File.join(node_modules_path, '.rnrepo-cache')
 
-        # Create Current directory if it doesn't exist (defaults to Debug, will be updated at build time)
+        # Create Current directory
         current_link = File.join(cache_dir, 'Current')
         FileUtils.rm_f(current_link) if File.exist?(current_link)
         # Prefer Debug for development
         debug_cache_dir = File.join(cache_dir, 'Debug')
-        release_cache_dir = File.join(cache_dir, 'Release')
-        default_config = File.exist?(debug_cache_dir) ? 'Debug' : 'Release'
-        FileUtils.cp_r(File.join(cache_dir, default_config), current_link)
-        CocoapodsRnrepo::Logger.log "  Copied to Current directory -> #{default_config}"
+        FileUtils.cp_r(debug_cache_dir, current_link)
+        CocoapodsRnrepo::Logger.log "  Copied to Current directory from Debug"
 
         # Look for xcframeworks in Current (which is a copy of Debug or Release)
         xcframeworks = Dir.glob(File.join(current_link, "*.xcframework"))
@@ -305,7 +303,7 @@ module Pod
             spec.attributes_hash[platform] ||= {}
 
             # Use simple relative path from pod directory
-            xcframework_relative_path = '.rnrepo-cache/Current/' + xcframework_name
+            xcframework_relative_path = File.join('.rnrepo-cache', 'Current', xcframework_name)
 
             # Add static xcframework as vendored_frameworks
             # CocoaPods handles static xcframeworks automatically


packages/build-tools/cocoapods-plugin/lib/plugin.rb
```diff
@@ -233,8 +233,8 @@ class << self
           debug_cache_dir = File.join(cache_dir, 'Debug')
           release_cache_dir = File.join(cache_dir, 'Release')
           default_config = File.exist?(debug_cache_dir) ? 'Debug' : 'Release'
-          FileUtils.ln_s(default_config, current_link)
-          CocoapodsRnrepo::Logger.log "  Created Current symlink -> #{default_config}"
+          FileUtils.cp_r(File.join(cache_dir, default_config), current_link)
```
kmagiera: we should watch out for the case when file already exists as it could be a symlink. Wonder if that isn't going to break the cocoapods logic that would resolve the link

packages/build-tools/cocoapods-plugin/lib/plugin.rb
```diff
@@ -305,15 +305,8 @@ class << self
...
-            # Create a symlink at pod root pointing to the Current framework
-            # This allows CocoaPods to find it easily
-            pod_root_link = File.join(node_modules_path, xcframework_name)
-            unless File.exist?(pod_root_link)
-              FileUtils.ln_s(File.join('.rnrepo-cache', 'Current', xcframework_name), pod_root_link)
-            end
-
             # Use simple relative path from pod directory
-            xcframework_relative_path = xcframework_name
+            xcframework_relative_path = '.rnrepo-cache/Current/' + xcframework_name
```
kmagiera: this should be enough to get around the symlink check in cocoapods. `.rnrepo-cache/Current/Name.xcframework` is not a symlink, only `.rnrepo-cache/Current` is

packages/build-tools/cocoapods-plugin/lib/plugin.rb
```diff
@@ -413,27 +376,46 @@ def rnrepo_post_install(installer_context)
...
 
-    # Move the script phase to run before "Compile Sources" phase
-    compile_phase = pod_target.source_build_phase
-    if compile_phase
-      pod_target.build_phases.delete(script_phase)
-      compile_phase_index = pod_target.build_phases.index(compile_phase)
-      pod_target.build_phases.insert(compile_phase_index, script_phase)
+  # Get list of prebuilt pod names for filtering
+  prebuilt_pod_names = (Pod::Installer.prebuilt_rnrepo_pods || []).map { |pod| pod[:name] }
+  script_name = '[AA] RNREPO Build Start'
```
kmagiera: add a comment here explaining why it needs to start with `[A`, or perhaps rename `[AAA Run First]` for more clarity

packages/build-tools/cocoapods-plugin/lib/plugin.rb
```diff
@@ -413,27 +376,46 @@ def rnrepo_post_install(installer_context)
...
+    next unless matches_prebuilt_pod
+
+    # Remove any existing RNREPO build phase first
+    target.build_phases.select { |phase| phase.is_a?(Xcodeproj::Project::Object::PBXShellScriptBuildPhase) && phase.name == script_name }.each(&:remove_from_project)
+    
+    # Create shell script build phase
+    build_phase = target.new_shell_script_build_phase(script_name)
+    build_phase.shell_script = script_content
+    
+    # Move it to the beginning by reordering build phases
```
kmagiera: this comment is not accurate. The phase will be added at the start but that won't actually make it appear as first unless the name is first alphabetically

packages/build-tools/cocoapods-plugin/lib/plugin.rb
radoslawrolka: done

packages/build-tools/cocoapods-plugin/lib/plugin.rb
radoslawrolka: it is not enough, the symlink must be somewhere else resolved to realpath

packages/build-tools/cocoapods-plugin/lib/plugin.rb
radoslawrolka: done

packages/build-tools/cocoapods-plugin/lib/plugin.rb
radoslawrolka: removed this code, as it is useless due to the alphabetical run order

packages/build-tools/cocoapods-plugin/lib/plugin.rb
```diff
@@ -226,18 +226,17 @@ class << self
         # Find the xcframework (name may differ from pod name due to sanitization)
         cache_dir = File.join(node_modules_path, '.rnrepo-cache')
 
-        # Create Current symlink if it doesn't exist (defaults to Debug, will be updated at build time)
+        # Create Current directory if it doesn't exist (defaults to Debug, will be updated at build time)
```
kmagiera: this comment it outdated, we remove the directory if symlink exists

packages/build-tools/cocoapods-plugin/lib/plugin.rb
```diff
@@ -226,18 +226,17 @@ class << self
...
-          release_cache_dir = File.join(cache_dir, 'Release')
-          default_config = File.exist?(debug_cache_dir) ? 'Debug' : 'Release'
-          FileUtils.ln_s(default_config, current_link)
-          CocoapodsRnrepo::Logger.log "  Created Current symlink -> #{default_config}"
-        end
+        FileUtils.rm_f(current_link) if File.exist?(current_link)
+        # Prefer Debug for development
+        debug_cache_dir = File.join(cache_dir, 'Debug')
+        release_cache_dir = File.join(cache_dir, 'Release')
+        default_config = File.exist?(debug_cache_dir) ? 'Debug' : 'Release'
```
kmagiera: should we signal somehow the situation when Debug doesn't exist instead of falling back to Release? Is it acceptable that Release exists while Debug doesn't?

packages/build-tools/cocoapods-plugin/lib/plugin.rb
```diff
@@ -305,15 +304,8 @@ class << self
...
-            # Create a symlink at pod root pointing to the Current framework
-            # This allows CocoaPods to find it easily
-            pod_root_link = File.join(node_modules_path, xcframework_name)
-            unless File.exist?(pod_root_link)
-              FileUtils.ln_s(File.join('.rnrepo-cache', 'Current', xcframework_name), pod_root_link)
-            end
-
             # Use simple relative path from pod directory
-            xcframework_relative_path = xcframework_name
+            xcframework_relative_path = '.rnrepo-cache/Current/' + xcframework_name
```
kmagiera: we use File.join elsewhere, would be cleaner to stick to one convention

kmagiera (approved):
Left few inline comments. Please address these prior to merging

packages/build-tools/cocoapods-plugin/lib/plugin.rb
radoslawrolka: done

packages/build-tools/cocoapods-plugin/lib/plugin.rb
radoslawrolka: It is not acceptable, was fixed in #250 - must both configuration exist to properly substitute lib

packages/build-tools/cocoapods-plugin/lib/plugin.rb
radoslawrolka: done

PR #250, opened 2026-02-10
Remove symlink to missing source

radoslawrolka:
## 📝 Description

Previously if Debug/release was missing then the present artifact was symlinked to the missing dir - that caused failed builds, so it was unnecessary.
Now framework_cache will not create symlink and return :unavailable with proper log message

And refactor of duplicated code in downloader

## 🎯 Type of Change

### DIFF
--- a/packages/build-tools/cocoapods-plugin/lib/framework_cache.rb
+++ b/packages/build-tools/cocoapods-plugin/lib/framework_cache.rb
@@ -39,37 +39,8 @@ module CocoapodsRnrepo
       # Matches sanitizePackageName() in @rnrepo/config
       sanitized_name = npm_package_name.gsub(/^@/, '').gsub('/', '_')
 
-      # First, check if both configurations exist on the server before downloading
-      Logger.log "Checking if both Debug and Release configurations are available on server..."
-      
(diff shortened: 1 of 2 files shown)


packages/build-tools/cocoapods-plugin/lib/framework_cache.rb
```diff
@@ -55,13 +84,11 @@ def self.fetch_framework(installer, pod_info, workspace_root)
           pod_name,
           worklets_version
         )
-        if debug_result[:status] == :downloaded || debug_result[:status] == :cached
-          configs_available << 'Debug'
-        else
-          Logger.log "⚠️  Debug configuration not available"
+        unless debug_result[:status] == :downloaded
+          Logger.log "Failed to download/extract Debug configuration"
+          cleanup_cache_dir(cache_dir)
```
kmagiera: do we need to cleanup actually? If only one flavor (debug or release) fails, we will save on sending head request later because we will have the file downloaded.

radoslawrolka: removed cleanup

PR #254, opened 2026-02-11
builder - cache rn project

radoslawrolka:
## 📝 Description

Create caching RN project in builder for each rn_version and OS.
- added in yamls creating clean RN project (nodes installed, no build artifacts from previous builds)
  - if work-dir provided to script already have rnrepo_build_app then its cached project (always happens on CI)
  - if work-dir provided to script does not have rnrepo_build_app then it creates one (with same script as CI creating cached one)
- added `createReactNativeProject` with
  - simple App.tsx (just view)
  - removed installed-by-default safe-area-context

## 🎯 Type of Change

- [x] ⚡ Performance improvement

## 🧪 Testing

<del>

- create cache in first build
  - workflow: https://github.com/software-mansion/rnrepo/actions/runs/21900137272 
  - cache: https://github.com/software-mansion/rnrepo/actions/caches#:~:text=Linux%2Drn%2Dproject%2Dcache%2D0.80.0
- use that cache in another build
  - workflow: https://github.com/software-mansion/rnrepo/actions/runs/21900317992
</del>

- create cache in first build
  - workflow: https://github.com/software-mansion/rnrepo/actions/runs/24077312315
- use that cache in another build
  - workflow: https://github.com/software-mansion/rnrepo/actions/runs/24077495031

### DIFF
--- a/packages/builder/build-library-android.ts
+++ b/packages/builder/build-library-android.ts
@@ -105,7 +105,7 @@ async function buildAAR(appDir: string, license: AllowedLicense) {
   }
 
   try {
-    await $`./gradlew :${gradleProjectName}:publishToMavenLocal \
+    await $`./gradlew ${gradleProjectName}:publishToMavenLocal \
       --no-daemon \
       --init-script ${addPublishingGradleScriptPath} \
       --init-script ${addPrefabReduceGradleScriptPath} \


packages/builder/build-utils.ts
```diff
@@ -227,6 +227,54 @@ export async function installSetup(
...
+
+export async function createReactNativeProject(
+  workDir: string, 
+  reactNativeVersion: string
+): Promise<void> {
+  console.log(
+    `📱 Creating temporary React Native project (RN ${reactNativeVersion})...`
+  );
+  // Create work directory if it doesn't exist
+  const { mkdirSync } = await import('fs');
```
WoLewicki: why are we doing inline imports instead of top level?

radoslawrolka: it was copied from other pr (codegen)

PR #280, opened 2026-03-05
Revert: "schedule 50 builds per scheduler cron #279"

radoslawrolka:
## 📝 Description

Revert of #279
Merge when we have custom macos runner

## 🎯 Type of Change

- [x] ⚡ Performance improvement


radoslawrolka:
This doesnt work (diff empty, creating new pr with reverting limiting 50 builds #308)

PR #281, opened 2026-03-05
emc worklets fix for exposdk@55

radoslawrolka:
### iOS

in exposdk@55 ExpoModulesCore direct dependency on RNWorklets was added, which resulted in:
```
❌  (node_modules/expo-modules-core/ios/Worklets/WorkletJSCallInvoker.h:10:10)

   8 | #include <ReactCommon/RuntimeExecutor.h>
   9 |
> 10 | #include <worklets/WorkletRuntime/WorkletRuntime.h>
     |          ^ 'worklets/WorkletRuntime/WorkletRuntime.h' file not found
  11 |
  12 | #include <memory>
  13 |
```
To solve this we have to propagate info about modulemap and headers dir to the ExpoModulesCore target. Adding it every time to EMC (without looking on version) does not break anything, because this info is not used without worklets dependency - also tested it on expo@54.

### Android

- expo@55 has [hardcoded worklets path in cmake file](https://github.com/expo/expo/blob/b887d67bbe061ac1f75ebcd9d018218868600822/packages/expo-modules-core/android/cmake/main.cmake#L83), so it have to be excluded from those versions 
- added removing cpp consumers from supported packages when cpp dependency is in denylist/unavailable

```
> Task :expo-modules-core:buildCMakeDebug[arm64-v8a] FAILED
...
  C++ build system [build] failed while executing:
      /Users/radoslawrolka/Library/Android/sdk/cmake/3.22.1/bin/ninja \
        -C \
        /Users/radoslawrolka/<...>/my-app-CLEAN-0503/node_modules/expo-modules-core/android/.cxx/Debug/v2v51444/arm64-v8a \
        expo-modules-core
    from /Users/radoslawrolka/<...>/my-app-CLEAN-0503/node_modules/expo-modules-core/android
  ninja: error: '/Users/radoslawrolka/<...>/my-app-CLEAN-0503/node_modules/react-native-worklets/android/build/intermediates/cmake/debug/obj/arm64-v8a/libworklets.so', needed by 
'../../../../build/intermediates/cxx/Debug/v2v51444/obj/arm64-v8a/libexpo-modules-core.so', missing and no known rule to make it
```

## 🎯 Type of Change

## 🧪 Testing

EXPOSDK@55
- sdk@55 ios - works
- sdk@55 android - works

Regression test for EXPOSDK@54
- sdk@54 ios - works
- sdk@54 android - works

testing repro:
```
npx create-expo-app@latest --template default@sdk-55 my-app-rnrepo-sdk55
cd my-app-rnrepo-sdk55
# add to package.json dependencies with this branch checkout
# # "@rnrepo/expo-config-plugin": "<path>/rnrepo/packages/expo-config-plugin",
# # "@rnrepo/build-tools": "<path>/rnrepo/packages/build-tools",
# add to app.json plugins
# # "@rnrepo/expo-config-plugin",
npx expo install
npx expo prebuild
npx expo run android
# success
npx expo run ios
# success
```

### DIFF
--- a/packages/build-tools/cocoapods-plugin/lib/plugin.rb
+++ b/packages/build-tools/cocoapods-plugin/lib/plugin.rb
@@ -98,6 +98,9 @@ def rnrepo_pre_install(installer_context)
     end
   end
 
+  # Add expo pod to installer context for later use in post_install
+  Pod::Installer.expo_pod = rn_pods.find { |pod| pod[:name] == 'Expo' }
+
   # Download and cache pre-built frameworks in parallel
   threads = rn_pods.map do |pod_info|
     Thread.new do
@@ -189,8 +192,10 @@ module Pod
     # Store prebuilt pod info (hashes with :name, :package_root, etc.) as class variable
     class << self
       attr_accessor :prebuilt_rnrepo_pods
+      attr_accessor :expo_pod
     end
     self.prebuilt_rnrepo_pods = []
+    self.expo_pod = nil
 
     # Hook into resolve_dependencies to modify specs BEFORE project generation
     old_method = instance_method(:resolve_dependencies)
@@ -420,12 +425,11 @@ def rnrepo_post_install(installer_context)
   # So when worklets are prebuilt, we need to add the modulesmaps to ExpoModulesCore target.
   installer_context.pods_project.targets.each do |target|
     if target.name == 'ExpoModulesCore'
+      worklets_pod = Pod::Installer.prebuilt_rnrepo_pods.find { |pod| pod[:name] == 'RNWorklets' }
+      expoVersion = Pod::Installer.expo_pod ? Pod::Installer.expo_pod[:version] : '999.0.0'
+      break if !worklets_pod || Gem::Version.new(expoVersion) < Gem::Version.new('55.0.0')
 
-      # Find Worklets package in node_modules
-      worklets_dependency = installer_context.podfile.target_definition_list
-        .map { |t| t.dependencies.find { |d| d.name == 'RNWorklets' } }
-        .compact.first
-      worklets_root = File.expand_path(worklets_dependency.external_source[:path])
+      worklets_root = worklets_pod[:package_root]
       if worklets_root == nil
         raise "RNWorklets not found in podfile, add react-native-worklets to denyList."
       end
@@ -446,6 +450,8 @@ def rnrepo_post_install(installer_context)
         build_settings['OTHER_SWIFT_FLAGS'] ||= '$(inherited)'
         build_settings['OTHER_SWIFT_FLAGS'] += " -Xcc #{module_map}"
       end
+
+      break
     end
   end
 end


packages/build-tools/gradle-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
```diff
@@ -664,10 +664,59 @@ class PrebuildsPlugin : Plugin<Project> {
                     return false
                 }
             }
+            "react-native-worklets" -> {
+                // In expo@55 and later, react-native-worklets is a dependency of expo, with hardcoded libworklets.so path in cmake file.
```
WoLewicki, line 668: Are you sure it will be present in the later versions? (just a nitpick, we can leave it like this :D )

packages/build-tools/README.md
```diff
@@ -40,14 +40,13 @@ buildscript {
...
 
-apply plugin: "com.facebook.react.rootproject"
-
 allprojects {
   repositories {
 +    maven { url "https://packages.rnrepo.org/releases" }
   }
 }
 
+apply plugin: "com.facebook.react.rootproject"
```
WoLewicki, line 49: Why was this moved?

packages/build-tools/gradle-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
```diff
@@ -664,10 +664,59 @@ class PrebuildsPlugin : Plugin<Project> {
...
+                // https://github.com/expo/expo/blob/b887d67bbe061ac1f75ebcd9d018218868600822/packages/expo-modules-core/android/cmake/main.cmake#L83
+                // So until that will be resolved in expo, we need to deny substitution for react-native-worklets if expo@55 is present in the project.
```
WoLewicki, line 670: That sounds bad. Does it mean that `react-native-reanimated` cannot be prebuilt right now since it needs prebuilt worklets? Or is it the case only in debug?

packages/build-tools/gradle-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
radoslawrolka: I dont know, but it will be futureproof, if it will change in next sdk then we will change comment too

packages/build-tools/README.md
radoslawrolka: I have observed that sometimes there was something like race-condition, so some packages had no repositories to look up for artifacts, and some had. After this change it was not observed anymore (in every other place in docs/scripts we have apply after repositories definition)

packages/build-tools/gradle-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
radoslawrolka: That is bad, rn-reanimated can not use artifacts in current state, even in debug

PR #319, opened 2026-04-16
Add explicit worklets installation for reanimated@4.3.0+

radoslawrolka:
## 📝 Description

Pod install failed when worklets were installed as a peer dependency

I’d avoid tying this to the old `workletsVersion` approach in `libraries.json`, since that was a reanimated-specific addition for its build process. Now that reanimated has been decoupled from worklets, that field is likely to be removed in the future

## 🎯 Type of Change

### DIFF
--- a/postInstallScripts/react-native-reanimated/react-native-reanimated.ts
+++ b/postInstallScripts/react-native-reanimated/react-native-reanimated.ts
@@ -1,0 +1,43 @@
+import { $ } from 'bun';
+import semver from 'semver';
+
+const COMPATIBILITY_MATRIX = [
+  {
+    reanimated: '>=4.3.0 <4.4.0',
+    worklets: '0.8.1'
+  },
+];
+
+export default async function postInstallSetup(): Promise<void> {
+    console.log(`Running post-install setup for react-native-reanimated...`);
+    
+    const reanimatedVersion = await getPackageVersion('react-native-reanimated');
+    console.log(`Detected react-native-reanimated version: ${reanimatedVersion}`);
+    const workletsVersion = COMPATIBILITY_MATRIX.find(m => semver.satisfies(reanimatedVersion, m.reanimated))?.worklets;
+    if (!workletsVersion) {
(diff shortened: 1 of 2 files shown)


postInstallScripts/react-native-reanimated/react-native-reanimated.ts
```diff
@@ -0,0 +1,43 @@
+import { $ } from 'bun';
+import semver from 'semver';
+
+const COMPATIBILITY_MATRIX = [
+  {
+    reanimated: '>=4.3.0 <4.4.0',
```
WoLewicki: You'll have to update it each time new versions comes out? Isn't it better to leave `reanimated: '>=4.3.0'` and count that it works in future?

WoLewicki (approved):
Approving but think about it

postInstallScripts/react-native-reanimated/react-native-reanimated.ts
radoslawrolka: i wanted to reduce the failing builds in the future, but act-when-failure is good too - changed to just `>=4.3.0`

postInstallScripts/react-native-reanimated/react-native-reanimated.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export default async function postInstallSetup(): Promise<void> {
+    console.log(`Running post-install setup for react-native-reanimated...`);
+    
+    const reanimatedVersion = await getPackageVersion('react-native-reanimated');
+    console.log(`Detected react-native-reanimated version: ${reanimatedVersion}`);
+    const workletsVersion = COMPATIBILITY_MATRIX.find(m => semver.satisfies(reanimatedVersion, m.reanimated))?.worklets;
+    if (!workletsVersion && semver.satisfies(reanimatedVersion, '>=4.3.0')) {
+        throw new Error(`Unsupported react-native-reanimated version: ${reanimatedVersion}. Please check the compatibility matrix.`);
+    }
```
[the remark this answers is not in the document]
radoslawrolka: done

PR #327, opened 2026-04-28
lock the mvn cli version used in publisher

radoslawrolka:
## 📝 Description

In `mvn deploy-file`, the `-packaging` argument was not used in version 3.1.2 (bug) which was fixed in later versions (currently worklfow uses 3.1.4). apache/maven-deploy-plugin/issues/576

It caused the iOS artifacts to be deployed with various extensions:
- old and wrong: `<artifact-name>.zip`
- new and correct: `<artifact-name>.xcframework.zip` 

What have been done:
- lock the version of mvn:deploy-file to 3.1.4 and gpg to 3.2.8 to avoid issue mentioned above 
  - versions locked on are same as currently used
- updated all `.zip` uses to be `.xcframework.zip`

BREAKING CHANGE:
- RENAMED (reupload) all existing artifacts on maven that are named `^.*(?<!\.xcframework)\.zip$` to `.xcframework.zip`

## 🎯 Type of Change

## Testing

- builds:
  - android: 
    - build: https://github.com/software-mansion/rnrepo/actions/runs/25054559013
    - publish: https://github.com/software-mansion/rnrepo/actions/runs/25054727591
  - ios: 
    - build: https://github.com/software-mansion/rnrepo/actions/runs/25053999050
    - publish: https://github.com/software-mansion/rnrepo/actions/runs/25054114005
  - artifacts: https://packages.rnrepo.org/snapshots/org/rnrepo/public/react-native-firebase_app/23.5.0

### DIFF
--- a/packages/build-tools/CHANGELOG.md
+++ b/packages/build-tools/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Worklets classifier is no longer added in reanimated@4.3.0+ artifacts (#317)
 - Incorporate RNRepo maven url inside gradle plugin (#270)
+- iOS plugin looks now for artifacts named '<>.xcframework.zip' (#327)
 
 ### Fixed
 - Added dummy header for iOS pods source_files (#311)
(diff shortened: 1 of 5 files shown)


kmagiera (approved):
Looks good, perhaps we could have the versions defined in some shared file such that it is easier to spot and update in the future

PR #336, opened 2026-05-11
Add claude.md

radoslawrolka:
## 📝 Description

Created via /init

> Note: wait until merged #335 or change this file

## 🎯 Type of Change

- [x] 📚 Documentation update

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,110 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## What this project is
+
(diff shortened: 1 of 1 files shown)


kmagiera (approved):
would be good to have some brief structure of modules (from packages/) and what they do to improve file searches when you ask the agent to do certain task that is typically limited to a single module scope.

Lets rename this to AGENTS.md for other agents to be able to pick up and remove claude code references from the content.

You can then create claude.md with the following content for claude to pick it up:
```
@AGENTS.md
```

radoslawrolka:
After thinking about it, I don’t think this file will be needed since the codebase isn’t that big and all informations are easily accessible through `package.json`, markdown files, and the existing structure. Usually the whole project context also isn’t required for everyday tasks.

Issue #334, opened 2026-05-06
[FEATURE]


ItsaMeTuni:
### 📝 Feature Description

Hi!

First and foremost, thanks for this awesome project!

Currently, the pre-built artifacts are stored in`.rnrepo-cache` folders inside each dependency folder in `node_modules`. I don't know if there's a specific reason for this, but it would be great if we could have `.rnrepo-cache` as a single folder in the project root or somewhere else.

The current approach prevents me from having a local cache of these prebuilt artifacts, because my build process looks something like this:
- create a temp dir for the build
- copy the project source to it
- install dependencies (yarn cache kicks in)
- generate some code/assets specific to that build
- run the build locally with eas <-- here rnrepo always downloads the prebuilt artifacts instead of retrieving them from a cache, since the node_modules folder is a "clean" folder

I won't get too deep into why the build process looks like this, but I need it to be this way so I can build a white-label app for multiple customers while sharing the same codebase.

It would be great if there were a way to store the artifacts in a single `.rnrepo-cache` folder. If I were able to specify the path to that folder in `app.config.ts` it would be even better since I could point many builds (and even separate projects) to the same folder.

I tried implementing a workaround by keeping track of the `.rnrepo-cache` folders in each build and synchronizing them to a single (persistent) folder before and after each build but I did something wrong and (IIRC) cocoapods complained about something getting corrupted. Will try again in the near future though.

Would this be a feasible feature?

radoslawrolka:
Hi,
Work in progress :building_construction:, created #338

ItsaMeTuni:
Awesome, @radoslawrolka! Thank you very much for this!

PR #338, opened 2026-05-13
Move xcframework cache to project level

radoslawrolka:
## 📝 Description

Relocate the primary cache directory from node_modules/.rnrepo-cache to:
- default:  ~/.rnrepo-cache/ 
- proviede in rnrepo.config.json["xcframeworksCacheDir"]: "/path/to/cache"
- disabled caching via rnrepo.config.json["xcframeworksCacheDir"]: "" | null

so it can be cached by runner CI.

#334 

## 🎯 Type of Change


packages/build-tools/cocoapods-plugin/lib/framework_cache.rb
```diff
@@ -14,19 +14,24 @@ def self.fetch_framework(installer, pod_info, workspace_root)
...
-      debug_cache_dir = File.join(cache_dir, 'Debug')
-      release_cache_dir = File.join(cache_dir, 'Release')
-
-      debug_exists = Dir.exist?(debug_cache_dir) && !Dir.glob(File.join(debug_cache_dir, "*.xcframework")).empty?
-      release_exists = Dir.exist?(release_cache_dir) && !Dir.glob(File.join(release_cache_dir, "*.xcframework")).empty?
-
-      if debug_exists && release_exists
+      if %w[Debug Release].all? { |c|
+           xcframework_exists?(File.join(cache_dir, c)) &&
+           (cache_dir_symlink.nil? || xcframework_exists?(File.join(cache_dir_symlink, c)))
```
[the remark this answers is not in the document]
radoslawrolka: done

packages/build-tools/cocoapods-plugin/README.md
```diff
@@ -73,3 +75,17 @@ node_modules/
...
+
+```bash
+RNREPO_CACHE_DIR=/path/to/shared/cache pod install
+```
+
+When this variable is set, artifacts are stored under `$RNREPO_CACHE_DIR/.rnrepo-cache/{package-name}/` and a symlink is created at `node_modules/{package-name}/.rnrepo-cache` pointing to that location so CocoaPods can still locate the frameworks. The layout inside is the same as above.
+
+This is particularly useful in CI environments where you can restore/save `$RNREPO_CACHE_DIR` between runs to avoid re-downloading unchanged frameworks.
+
+> **Switching cache modes:** If you switch between using `RNREPO_CACHE_DIR` and the default mode, existing cached directories in `node_modules` will be replaced with symlinks (or vice versa) automatically on the next `pod install`.
```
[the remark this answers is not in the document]
radoslawrolka: it works like was written now

packages/build-tools/cocoapods-plugin/lib/framework_cache.rb
```diff
@@ -14,19 +14,24 @@ def self.fetch_framework(installer, pod_info, workspace_root)
       node_modules_path = File.expand_path(source_path, workspace_root)
       Logger.log "Package directory: #{node_modules_path}"
 
-      # Create .rnrepo-cache directory
-      cache_dir = File.join(node_modules_path, '.rnrepo-cache')
+      # When RNREPO_CACHE_DIR is set, artifacts live there and symlinks are created
+      # in node_modules/<package>/.rnrepo-cache so CocoaPods can find them.
+      # Without it, artifacts are stored directly inside the package directory.
+      if ENV['RNREPO_CACHE_DIR']
+        cache_dir = File.join(File.expand_path(ENV['RNREPO_CACHE_DIR'], workspace_root), '.rnrepo-cache', pod_name)
```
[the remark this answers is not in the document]
radoslawrolka: done

packages/build-tools/cocoapods-plugin/README.md
```diff
@@ -73,3 +73,13 @@ node_modules/
...
           └── Current/  (symlink created at build time → Debug or Release)
 ```
+
+### Cache directory
+
+By default, the plugin downloads artifacts to `~/.rnrepo-cache` directory. This can be overridden by setting `iOSAllowHomeDirCache` to `false` in `rnrepo.config.json`.
+
+```json
+{
+  "iOSAllowHomeDirCache": false
```
kmagiera: I feel like it'd be more practical to just generalize it to accept a cache directory instead of YES/NO for the home dir. For disabling the cache entirely you could just set it to null/empty.

Same goes for the plugin code where we have a flag that tells whether to have it in homedir or not. This should really just be a cache dir path / null instead.

kmagiera (changes requested):

packages/build-tools/cocoapods-plugin/README.md
radoslawrolka: changed to:
**iOS:** Downloaded xcframeworks are extracted to `node_modules/{package-name}/.rnrepo-cache` on every `pod install` and cached in `~/.rnrepo-cache` by default. To use a custom cache path or disable caching, set `iOSCachePath` in `rnrepo.config.json`:

```json
{
  "iOSCachePath": "/path/to/cache"
}
```

To disable caching entirely, set `iOSCachePath` to `null`.

packages/build-tools/cocoapods-plugin/lib/plugin.rb
```diff
@@ -65,6 +65,14 @@ def get_ios_denylist(workspace_root)
   denylist_config['ios'] || []
 end
 
+def get_ios_cache_path(workspace_root)
+  config = load_rnrepo_config(workspace_root)
+  return File.expand_path('~/.rnrepo-cache') unless config.key?('iOSCachePath')
```
kmagiera: the path specified in the config file needs to be relative to the location of the config file (unless it is defined as absolute path with leading slash).

I'd also use another name for the config like `xcframeworksCacheDir`

radoslawrolka: done

packages/build-tools/cocoapods-plugin/README.md
```diff
@@ -73,3 +73,15 @@ node_modules/
...
           └── Current/  (symlink created at build time → Debug or Release)
 ```
+
+### Cache directory
+
+By default, downloaded artifacts are cached in `~/.rnrepo-cache`. To use a custom path, set `xcframeworksCacheDir` in `rnrepo.config.json`:
+
+```json
+{
+  "xcframeworksCacheDir": "/path/to/cache"
```
kmagiera, line 83: we should document that the path is relative to the .rnrepo-cache file location (ofc when it starts with / then it is absolute)

README.md
```diff
@@ -163,6 +163,20 @@ If you're unsure whether RNRepo has been set up correctly in your project, check
 
 For troubleshooting Android builds, before reporting an issue, we recommend passing the `--scan` flag to Gradle (e.g., `./gradlew app:assembleDebug --scan`). This flag generates a report of all tasks performed during the build along with their execution times, which can be useful for investigating issues such as when certain prebuilt libraries weren't loaded from the repository.
 
+### Artifact caching
+
+**Android:** The Gradle plugin relies on Gradle's built-in dependency cache. Downloaded `.aar` artifacts are stored automatically in `~/.gradle/caches/` and reused across builds and projects without any extra configuration. To warm the cache in CI, save and restore the `~/.gradle/caches/` directory between runs.
+
+**iOS:** Downloaded xcframeworks are extracted to `node_modules/{package-name}/.rnrepo-cache` on every `pod install` and cached in `~/.rnrepo-cache` by default. To use a custom cache path or disable caching, set `xcframeworksCacheDir` in `rnrepo.config.json`:
```
kmagiera: that's a weird sequence of explaining. They are first downloaded into .rnrepo-cache and then extracted, not the other way around

packages/build-tools/cocoapods-plugin/lib/plugin.rb
```diff
@@ -65,6 +69,16 @@ def get_ios_denylist(workspace_root)
   denylist_config['ios'] || []
 end
 
+def get_ios_cache_path(workspace_root)
+  config = load_rnrepo_config(workspace_root)
+  return File.expand_path('~/.rnrepo-cache') unless config.key?('xcframeworksCacheDir')
+  path = config['xcframeworksCacheDir']
+  return nil if !path || path.to_s.strip.empty?
+  return File.expand_path(path) if path.start_with?('/') || path.start_with?('~')
```
kmagiera: gpt says that expand_pah already handles abslute paths, so instead of doing all the checks you could just do:
```
config_dir = File.dirname(find_rnrepo_config_path(workspace_root))
return File.expand_path(path, config_dir)
```

kmagiera (approved):

packages/build-tools/cocoapods-plugin/README.md
radoslawrolka: done

README.md
radoslawrolka: done

packages/build-tools/cocoapods-plugin/lib/plugin.rb
radoslawrolka: done

packages/build-tools/cocoapods-plugin/lib/plugin.rb
```diff
@@ -65,6 +69,15 @@ def get_ios_denylist(workspace_root)
...
 end
 
+def get_ios_cache_path(workspace_root)
+  config = load_rnrepo_config(workspace_root)
+  return File.expand_path('~/.rnrepo-cache') unless config.key?('xcframeworksCacheDir')
+  path = config['xcframeworksCacheDir']
+  return nil if !path || path.to_s.strip.empty?
+  config_dir = File.dirname(find_rnrepo_config_path(workspace_root))
+  File.expand_path(path, config_dir)
+end
```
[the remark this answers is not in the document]
radoslawrolka: updated pr description

PR #364, opened 2026-06-08
fix rngh patch for 2.25.0 and +3.0.0

radoslawrolka:
## 📝 Description

2.25.0 and +3.0.0 builds failed due to patch applying failures. Only diffs in .patch files are single lines:
```
$ diff react-native-gesture-handler.patch react-native-gesture-handler_2.25.0.patch

62c62
< @@ -1,65 +1,14 @@
---
> @@ -1,67 +1,14 @@
95c95,97
< -    fun isSvgElement(view: Any): Boolean = (view is VirtualView || view is SvgView)
---
> -    fun isSvgElement(view: Any): Boolean {
> -      return (view is VirtualView || view is SvgView)
> -    }
```
and
```
$ diff react-native-gesture-handler.patch react-native-gesture-handler_3.0.0.patch

29c29
<
---
>
36,38c36,38
<
<              if (isNewArchitectureEnabled()) {
<                  srcDirs += 'fabric/src/main/java'
---
>          }
>      }
>

```


WoLewicki (approved):
This sounds like something we will have to constantly take care of yeah? Could we maybe somehow avoid that, cc @j-piasecki also.

PR #372, opened 2026-06-11
Fix intermittent "collector is null" NPE in debug builds (sequential C++ dependency check)

### LINKED ISSUE #371: [BUG] Intermittent `NullPointerException: Cannot invoke "Collectors$TypedCollector.collectInto(...)" because "collector" is null` on debug builds (race condition in `parallelStream()` dependency check)
⚠️ This was mostly generated by a LLM ⚠️ 
I edited it, and it seemed to mostly make sense to me but I do not know enough the topic to be sure.

## Description

Debug builds intermittently fail during the configuration phase (right after all projects are evaluated) with:

```
FAILURE: Build failed with an exception.

* What went wrong:
java.lang.NullPointerException (no error message)
> Cannot invoke "org.gradle.api.internal.provider.Collectors$TypedCollector.collectInto(com.google.common.collect.ImmutableCollection$Builder)" because "collector" is null
```

The failure is non-deterministic — in our project it reproduces on roughly 40% of cold-daemon runs of `./gradlew app:assembleDevDebug`, and never on release builds.

## Root cause

In `PrebuildsPlugin.apply`, the debug-only dependency compatibility check runs inside `gradle.projectsEvaluated` using a **`parallelStream()`**:

```kotlin
if (getBuildType(project) == "debug") {
    project.gradle.projectsEvaluated {
        PACKAGES_WITH_CPP.keys.parallelStream().forEach { packageName ->
            ...
            checkDependencies(packageItem, project, extension.supportedPackages)
        }
    }
}
```

`checkDependencies` then iterates every subproject's configurations and dependencies from each parallel thread:

```kotlin
project.rootProject.allprojects.forEach { subproject ->
    ...
    val dependsOnPackage =
        subproject.configurations.any { config ->
            config.dependencies.any { dependency ->
                dependency.name == packageItem.name
            }
        }
    ...
}
```

`Configuration.getDependencies()` is a lazy Gradle domain object collection backed by pending provider collectors, and these are **not thread-safe**. When several `parallelStream()` worker threads realize the same pending collectors concurrently, the collector list is occasionally corrupted and Gradle throws the NPE above. This is why the failure is intermittent, only affects debug builds (the check doesn't run otherwise), and is independent of the Gradle version (reproduced on Gradle 9.0.0 and 9.3.1).

## Stack trace

Captured with `--stacktrace` on plugin version 0.2.2 / Gradle 9.3.1 (line numbers match the `@rnrepo/build-tools@0.1.4` tag — `PrebuildsPlugin.kt:155` is the `projectsEvaluated` lambda, `:157` is the `parallelStream().forEach` body):

```
java.lang.NullPointerException
	at org.rnrepo.tools.prebuilds.PrebuildsPlugin$apply$4.invoke(PrebuildsPlugin.kt:157)
	at org.rnrepo.tools.prebuilds.PrebuildsPlugin$apply$4.invoke(PrebuildsPlugin.kt:155)
	at org.rnrepo.tools.prebuilds.PrebuildsPlugin.apply$lambda$7(PrebuildsPlugin.kt:155)
	at org.gradle.internal.code.DefaultUserCodeApplicationContext$CurrentApplication$1.execute(DefaultUserCodeApplicationContext.java:124)
	at org.gradle.configuration.internal.DefaultListenerBuildOperationDecorator$BuildOperationEmittingAction$1.run(DefaultListenerBuildOperationDecorator.java:173)
	...
	at jdk.proxy1/jdk.proxy1.$Proxy39.projectsEvaluated(Unknown Source)
	at org.gradle.initialization.ProjectsEvaluatedNotifier$NotifyProjectsEvaluatedListeners.run(ProjectsEvaluatedNotifier.java:47)
	at org.gradle.initialization.ProjectsEvaluatedNotifier.notify(ProjectsEvaluatedNotifier.java:35)
	at org.gradle.configuration.DefaultProjectsPreparer.prepareProjects(DefaultProjectsPreparer.java:52)
```

Full stacktrace here [rnrepo-issue-stacktrace.txt](https://github.com/user-attachments/files/28849145/rnrepo-issue-stacktrace.txt)

## Environment

- `org.rnrepo.tools:prebuilds-plugin`: **0.2.2** (latest)
- React Native: **0.83.0** (new architecture enabled, Hermes)
- Gradle: reproduced on **9.0.0** and **9.3.1**
- AGP: 8.12.0 (RN default), Kotlin 2.1.20
- macOS (Apple Silicon), JDK 17
- `org.gradle.parallel=true`, `org.gradle.caching=true`, configuration cache **not** enabled
- ~60 autolinked subprojects

## Steps to reproduce

1. React Native app with a reasonably large number of native modules and the prebuilds plugin applied.
2. Run repeatedly with a cold daemon:
   ```bash
   for i in 1 2 3 4 5; do ./gradlew --stop; ./gradlew app:assembleDebug; done
   ```
3. A run fails with the NPE within a few attempts (~40% of cold runs for us).

## Suggested fix

Any of:

- Replace `parallelStream()` with a sequential `forEach` — the check iterates in-memory collections and is not performance-critical;
- or snapshot each subproject's dependencies on the calling thread first (e.g. `config.dependencies.toList()` single-threaded), then run the matching logic in parallel over the snapshot.

## Workaround we're using

Realizing all dependency collections single-threaded before the plugin's `projectsEvaluated` callback runs (registered **before** `apply plugin` so it executes first):

```gradle
// app/build.gradle
gradle.projectsEvaluated {
    rootProject.allprojects.each { p ->
        p.configurations.each { c -> c.dependencies.size() }
    }
}
apply plugin: "org.rnrepo.tools.prebuilds-plugin"
```

With this in place: 8/8 cold-daemon builds pass (vs ~60% before).

Thanks for what you do with RNRepo!
The build-time savings are huge for us!

gadcam:
⚠️ This was generated by a LLM ⚠️
## Summary

Fixes #371.

On debug builds, the C++ dependency compatibility check registered in `gradle.projectsEvaluated` iterated `PACKAGES_WITH_CPP.keys` with a `parallelStream()`. `checkDependencies` then realizes each subproject's `configurations`/`dependencies` from those worker threads — but Gradle's lazy dependency collections (pending provider collectors) are not thread-safe to realize concurrently. When two threads raced, the configuration phase failed intermittently with:

```
java.lang.NullPointerException (no error message)
> Cannot invoke "org.gradle.api.internal.provider.Collectors$TypedCollector.collectInto(com.google.common.collect.ImmutableCollection$Builder)" because "collector" is null
```

In our app (RN 0.83, ~60 autolinked subprojects) this hit roughly 40% of cold-daemon debug builds, on both Gradle 9.0.0 and 9.3.1.

## Fix

Replace `parallelStream()` with a sequential `forEach`. `PACKAGES_WITH_CPP` has two entries and each check only iterates in-memory dependency declarations, so the parallelism had no measurable benefit (the whole check is a few milliseconds) while introducing the race. A comment marks the loop as intentionally sequential.

## Testing

- `./gradlew test` in `packages/build-tools/gradle-plugin` — passes
- `./gradlew spotlessCheck` — passes
- Real-world verification on the app from #371: published the patched plugin to mavenLocal as `0.2.3-fix371` and ran 8 cold-daemon (`./gradlew --stop` between runs) `app:assembleDevDebug` builds with no other workaround in place: **8/8 passed**, with RNRepo prebuilds confirmed active in the logs. On plugin 0.2.2 the same loop failed ~40% of runs with the NPE above.

### DIFF
--- a/packages/build-tools/CHANGELOG.md
+++ b/packages/build-tools/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Resolve worklets framework module map (#352)
 - Detect debug configuration using GCC preprocessor definitions (#367)
+- Run C++ dependencies check sequentially instead of in parallel (#372)
 
 ## [0.1.4] - 2026-05-21
 
--- a/packages/build-tools/gradle-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
+++ b/packages/build-tools/gradle-plugin/src/main/kotlin/org/rnrepo/tools/prebuilds/PrebuildsPlugin.kt
@@ -153,7 +153,11 @@ class PrebuildsPlugin : Plugin<Project> {
             if (getBuildType(project) == "debug") {
                 project.gradle.projectsEvaluated {
                     logger.info("Checking if all dependencies with c++ code have their consumers supported...")
-                    PACKAGES_WITH_CPP.keys.parallelStream().forEach { packageName ->
+                    // Must stay sequential: checkDependencies iterates the subprojects' lazy
+                    // dependency collections, which are not thread-safe to realize concurrently
+                    // (intermittent "Collectors$TypedCollector.collectInto(...) collector is null"
+                    // NPE). See https://github.com/software-mansion/rnrepo/issues/371
+                    PACKAGES_WITH_CPP.keys.forEach { packageName ->
                         val packageItem = extension.projectPackages.find { it.name == packageName }
                         if (packageItem == null) {
                             logger.info("Package $packageName not found in project packages, skipping dependency check.")


radoslawrolka (approved):
lgtm, this does not introduce any additional time overhead.

PR #377, opened 2026-06-12
simplify issue reports templates

radoslawrolka:
## 📝 Description

Many unuseful info and very annoying to fill when reporting

### DIFF
--- a/.github/ISSUE_TEMPLATE/library_support_request.yml
+++ b/.github/ISSUE_TEMPLATE/library_support_request.yml
@@ -13,8 +13,13 @@ body:
     id: library-name
     attributes:
       label: 📚 Library Name
-      description: What library should be added to RNRepo?
-      placeholder: Library name here...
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/library_support_request.yml
```diff
@@ -35,22 +40,9 @@ body:
     attributes:
       label: GitHub Repository
       placeholder: "https://github.com/..."
-    validations:
```
WoLewicki, line 38: Can we build non-github libs?

WoLewicki (approved):

.github/ISSUE_TEMPLATE/library_support_request.yml
radoslawrolka: We can, but I’m not aware of any 100k+ downloads/week library that isn’t hosted on GH. Also link to npm is enough and usually have linked GH repo.

PR #406, opened 2026-07-06
Rolkrado/revert disable screens temporary

radoslawrolka:
## 📝 Description

reverting the #405


radoslawrolka:
DIFF DOESNT WORK - contimue in #407

PR #417, opened 2026-07-16
update npm version to 24

radoslawrolka:
## 📝 Description

Due to failing jobs:
- https://github.com/software-mansion/rnrepo/actions/runs/29537171571/job/87751132245


WoLewicki (approved):
I'd think if we could automate the process of taking the "state of the art" versions automatically.

PR #427, opened 2026-07-27
bump rngh

radoslawrolka:
## 📝 Description

- changed versions according to docs that has changed https://docs.swmansion.com/react-native-gesture-handler/docs/fundamentals/installation/#compatibility-with-react-native-versions

### DIFF
--- a/libraries.json
+++ b/libraries.json
@@ -25,8 +25,11 @@
       },
       {
         "versionMatcher": ">=3.0.0",
-        "reactNativeVersion": ">=0.82 <0.86",
-        "publishedAfterDate": "2025-01-01"
+        "reactNativeVersion": ">=0.82 <0.87"
+      },
+      {
+        "versionMatcher": ">=3.1.0",
+        "reactNativeVersion": ">=0.83 <0.87"
       }
     ],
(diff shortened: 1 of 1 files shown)


libraries.json
```diff
@@ -36,8 +39,11 @@
       },
       {
         "versionMatcher": ">=3.0.0",
-        "reactNativeVersion": ">=0.82 <0.86",
-        "publishedAfterDate": "2025-01-01"
+        "reactNativeVersion": ">=0.82 <0.87"
+      },
+      {
+        "versionMatcher": ">=3.1.0",
+        "reactNativeVersion": ">=0.83 <0.87"
```
WoLewicki: why do we do upper limit of RN versions? We assume they wont work ?

WoLewicki (approved):

libraries.json
radoslawrolka: before the 0.87 i wanted to test the core libs (rea, rngh, svg, screens) since the spm will be introduced there to avoid removing&rebuilding. But we have min-release-age=4days so that should be enough.

removed upperbound in those entries

PR #429, opened 2026-07-27
Keep library header overrides on the prebuilt codegen include path

radoslawrolka:
## 📝 Description

Apps consuming the react-native-gesture-handler prebuilt fail to compile with:

    autolinking.cpp:52: error: unknown type name
    'RNGestureHandlerDetectorComponentDescriptor'

RNGH 3.x ships a hand-written header that shadows its codegen output

#430 

### `packages/builder/gradle_init_scripts/add-publishing.gradle`

- Added `findCodegenOverrideIncludeRoots()`, which scans the packaged `assets/headers` tree for subdirectories carrying their own `react/renderer/components/<codegenName>/*.h` and returns them sorted, recovering the include-root information that flattening destroys.
- Added `findPackageCodegenOverrideRoots()`, which applies the same shape detection to the package sources to find override roots that lie outside every directory `packageCppHeaders` collects, skipping `build`/`node_modules`/`.cxx`/`.gradle` paths and any root already covered by a collected directory.
- `packageCppHeaders` now copies those extra roots into `assets/headers/<package-relative path>`, preserving the prefix instead of flattening them so they shadow rather than clobber the generated headers.
- `packageCppHeaders` now deletes `assets/headers` before repopulating it, so leftovers from an earlier run are neither packaged into the AAR nor mistaken for override include roots.
- `generateCMakeContent()` now takes the discovered override roots and emits them in `target_include_directories` ahead of `assets/headers`, mirroring the include order the library's own CMakeLists uses.
- `packageCodegenCMake` now declares `dependsOn(packageHeadersTask)`, since it derives the include roots from the directory that task fills.
- The `assets/headers` path is now a single `headersDestDir` variable shared by both tasks instead of a literal repeated in each.

### `packages/builder/build-library-android.ts`

- Extracted the `SHARED → STATIC` rewrite into a `CMAKE_STATIC_PATCHES` table and added an `OBJECT → STATIC` rule scoped to `react_codegen_*` targets, so hand-written codegen CMakeLists that declare an OBJECT library (react-native-reanimated) produce a linkable archive.
- Relaxed the early-exit precondition from "file contains `add_library` and `SHARED`" to just "file contains `add_library`", so OBJECT-only files are no longer skipped before patching.
- `patchCMakeListsToStatic()` now takes `hasCodegenConfig` and rethrows patch failures as fatal for libraries that publish a codegen AAR, instead of always warning and continuing.
- The success log now names which rewrites were applied rather than assuming `SHARED → STATIC`.

> note reanimated@4.5.0 with REASharedTransitionBoundaryComponentDescriptor work too

## 🧪 Testing

- AwesomeProject_86 (RN 0.86, new arch) build with new artifacts builds correctly
  - with rngh@+3 and reanimated@+4.5
- artifacts will be redeployed with proper codegen

### DIFF
--- a/packages/builder/gradle_init_scripts/add-publishing.gradle
+++ b/packages/builder/gradle_init_scripts/add-publishing.gradle
@@ -23,13 +23,44 @@ allprojects { project ->
     }
 }
 
-def generateCMakeContent(codegenName) {
+// Some libraries ship hand-written headers that deliberately shadow their codegen output. The
+// clearest case is react-native-gesture-handler: shared/shadowNodes holds its own copy of
+// react/renderer/components/rngesturehandler_codegen/ComponentDescriptors.h and only that copy
+// declares RNGestureHandlerDetectorComponentDescriptor. The library's own CMakeLists puts that
+// directory on the include path *before* the codegen one, so the override wins.
+//
+// packageCppHeaders flattens every header source into assets/headers, which loses the knowledge of
+// which subdirectories were include roots. Recover it here: any directory below assets/headers that
+// carries its own react/renderer/components/<codegenName> subtree is such an override root and has
+// to be searched before assets/headers itself, otherwise the plain codegen header wins and the
+// consuming app fails to compile.
+def findCodegenOverrideIncludeRoots(headersDir, codegenName) {
+    if (codegenName == null || !headersDir.isDirectory()) {
+        return []
+    }
+    def roots = []
+    // eachDirRecurse skips headersDir itself, which is what we want — the codegen tree lives
+    // directly under it and is emitted separately.
+    headersDir.eachDirRecurse { dir ->
+        if (new File(dir, "react/renderer/components/${codegenName}").isDirectory()) {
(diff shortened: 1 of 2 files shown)


packages/builder/gradle_init_scripts/add-publishing.gradle
```diff
@@ -23,13 +23,129 @@ allprojects { project ->
...
+    }
+    def headersPath = headersDir.toPath()
+    def marker = "react/renderer/components/${codegenName}/"
+    def roots = [] as Set
+
+    // Keyed off actual header files rather than directory existence: Gradle's copy recreates the
+    // whole directory structure even when the "**/*.h" filter matches nothing, so a subtree built
+    // solely from .cpp files would otherwise be reported as a root. Files.walk does not follow
+    // symlinks, so a link back to an ancestor cannot recurse forever.
+    def stream = java.nio.file.Files.walk(headersPath)
```
WoLewicki, line 56: shouldn't we just add it as an import a the top of the file?

WoLewicki (approved):

PR #441, opened 2026-08-26
add robots.txt, sitemap, llms.txt and structured data

halskiszymon:
rnrepo.org has no robots.txt of its own - cloudflare currently answers with a generic content-signals stub - and no sitemap.

- **robots.txt** with `Content-Signal: search=yes, ai-input=yes, ai-train=yes`, the signal we use across swmansion domains, plus a sitemap reference
- **sitemap.xml** listing both pages
- **llms.txt** - plain-text index for ai crawlers
- **json-ld** carrying the shared `https://swmansion.com/#organization` id, so rnrepo.org resolves to the same entity as the rest of our sites
- **canonical link**, which `BaseLayout` did not emit at all
- **/supported-libraries gets its own title and description** - it was inheriting the homepage's, so both pages looked identical to search engines

no new dependencies; sitemap is a route rather than `@astrojs/sitemap` since there are two pages. happy to switch if you'd rather have the integration.

check: `/robots.txt`, `/sitemap.xml`, `/llms.txt`, and the `application/ld+json` block in page source.


radoslawrolka (approved):
lgtm

PR #445, opened 2026-09-11
feat: migrate website to @swmansion/ui-components and Panda CSS

p-malecki:
## 📝 Description

Rebuilds the RNRepo website on top of the shared `@swmansion/ui-components`
design system, backed by Panda CSS, so the site stops maintaining its own
parallel copy of Software Mansion's typography, spacing, colour and button
styles. Every visual decision that used to live in Tailwind classes now
comes from library primitives and design tokens, which means the next brand or
component update arrives by bumping a dependency instead of hand-editing markup.

### Changes

- Migrated the website to `@swmansion/ui-components` + Panda CSS
- Rebuilt Hero, Features, How It Works, Custom Setup, FAQ, Getting Started, navbar and footer on library components
- Rebuilt the Supported Libraries page header and CTA the same way
- Added scroll reveal animations to every section
- Configured Panda with RNRepo's palette. Tailwind stays for the not-yet-migrated parts
- No copy, data or behaviour changes

## 🧪 Testing

Ran `npm run build`. It is clean and both pages were checked in the browser at 1900, 1600,
1440, 1280, 1024, 768 and 390px.

PR #446, opened 2026-09-11
chore: pass GL_TOKEN to dependency installs in CI

p-malecki:
## 📝 Description

In #445 the website is moving onto `@swmansion/ui-components`, which is published to the
private GitLab package registry. `.npmrc` maps the `@swmansion`
scope to that registry and reads the auth token from `${GL_TOKEN}`, but no
workflow sets that variable today, so CI installs unauthenticated and the
registry answers 401:

error: GET https://gl.swmansion.com/api/v4/projects/682/packages/npm/@swmansion%2fui-components - 401
error: @swmansion/ui-components@^1.0.59 failed to resolve in lint-and-validate

This adds `GL_TOKEN` to every step that runs the workspace install.

All 13 install steps are covered, not just `lint-and-validate` - each one runs
the root bun install, which resolves packages/website too, so every workflow
would hit the same 401. 

⚠️ Someone with admin access needs to add a `GL_TOKEN` repository secret. I got the token, but cannot add it without repo permissions.

⚠️ This PR should be merged before #445.

🧪 Testing

1. Add the `GL_TOKEN` secret.
2. Re-run CI on the website migration PR — `bun install` resolves
   `@swmansion/ui-components` instead of failing with a 401.

### DIFF
--- a/.github/workflows/build-library-android.yml
+++ b/.github/workflows/build-library-android.yml
@@ -52,6 +52,8 @@ jobs:
 
       - name: 📦 Install dependencies
         run: bun install --ignore-scripts
+        env:
+          GL_TOKEN: ${{ secrets.GL_TOKEN }}
 
       - name: ♻️ Restore RN project cache
         id: rn-project-cache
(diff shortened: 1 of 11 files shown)


radoslawrolka (approved):

radoslawrolka:
Closing in favour of #445, cherry-picked this commit to the #445 to keep all changes in one place

radoslawrolka:
Added secret GL_TOKEN

